PR #157, opened 2019-12-09
Feature/leases query parameters

nathanagood:
## Proposed changes

Added new router and refactored for supporting query parameters for leases REST API

## Types of changes

## Checklist


cmd/lambda/accounts/main.go
```diff
@@ -187,68 +185,3 @@ func newAWSSession() *session.Session {
 	}
 	return awsSession
 }
-
-// WriteServerErrorWithResponse - Writes a server error with the specific message.
-func WriteServerErrorWithResponse(w http.ResponseWriter, message string) {
```
joshmarsh: 👍

cmd/lambda/leases/get.go
```diff
@@ -1,39 +1,139 @@
...
+	leases, err := dao.FindLeasesByAccount(accountID)
+	if err != nil {
+		errMsg := fmt.Sprintf("Error getting leases for account %s: %s", accountID, err.Error())
+		log.Println(errMsg)
+		response.WriteServerErrorWithResponse(w, errMsg)
+		return
+	}
+
+	if len(leases) == 0 {
+		// We were throwing an error on these, but not sure that's the right thing
```
marinatedpork: Pardon me if I am over explaining here - given that it seems you've thought about this yourself via this comment, but a query with no results should not throw an HTTP error, but rather return a status 200 with an "empty" (like a JSON array without an elements) body. 404s are used for resources in path parameters that are not found. In the case of querying a collection with zero results, there really wasn't an error. If you query, nothing breaks, and you have a result empty result set, then you still have the correct results, which in this case happen to be empty.

Conclusion here being that this should be change to 200 OK with empty array as body.

cmd/lambda/leases/get.go
```diff
@@ -1,39 +1,139 @@
...
+
+	for _, l := range leases {
+		leaseResponse := response.LeaseResponse(*l)
+		leaseResponses = append(leaseResponses, &leaseResponse)
+	}
+
+	json.NewEncoder(w).Encode(leaseResponses)
+}
+
+// GetLeasesByStatus - Returns a list of leases by principal and account
```
marinatedpork: This method documentation seems incorrect

cmd/lambda/leases/handlers.go
```diff
@@ -0,0 +1,71 @@
...
+	// Iterate through the tasks. Pay attention to the failOnError
+	for _, t := range tasker.taskConfiguration {
+		err := t.handler(context, lease)
+		if err != nil {
+			tasker.errorList = append(tasker.errorList, fmt.Errorf("%s: %s", t.name, err.Error()))
+			if t.failOnError {
+				return false, err
+			}
+		}
+		// otherwise, nothing to see here... move along.
```
marinatedpork: ?

cmd/lambda/leases/validate.go
```diff
@@ -42,39 +51,39 @@ func validateLeaseRequest(controller CreateController, req *events.APIGatewayPro
...
-	usageRecords, err := controller.UsageSvc.GetUsageByDateRange(usageStartTime, usageEndTime)
-	if err != nil {
-		errStr := fmt.Sprintf("Failed to retrieve usage: %s", err)
-		return requestBody, true, "", errors.New(errStr)
-	}
+	// usageRecords, err := UsageSvc.GetUsageByDateRange(usageStartTime, usageEndTime)
+	// if err != nil {
+	// 	errStr := fmt.Sprintf("Failed to retrieve usage: %s", err)
+	// 	return requestBody, true, "", errors.New(errStr)
+	// }
```
marinatedpork: We want to keep the validation of leases in here.

marinatedpork: Or is there a different way we're validating them that I am not seeing?

cmd/lambda/leases/create.go
```diff
@@ -111,17 +106,18 @@ func (c CreateController) Call(ctx context.Context, req *events.APIGatewayProxyR
...
 		log.Printf("Failed to create lease DB record for %s @ %s: %s",
 			requestBody.PrincipalID, account.ID, err)
-		return response.ServerError(), nil
+		response.WriteServerError(w)
+		return
 	}
 
-	// Mark the account as Status=Leased
-	account, err = c.Dao.TransitionAccountStatus(account.ID, db.Ready, db.Leased)
+	// Mark the accOunt as Status=Leased
```
marinatedpork: typo. not a blocker, though. just bringing it to your attention

cmd/lambda/leases/main.go
```diff
@@ -34,74 +61,117 @@ type messageBody struct {
...
+			CreateLease,
+		},
+	}
+	r := api.NewRouter(leasesRoutes)
+	muxLambda = gorillamux.New(r)
+}
+
+// Handler - Handle the lambda function
+func Handler(ctx context.Context, req events.APIGatewayProxyRequest) (events.APIGatewayProxyResponse, error) {
+	// If no name is provided in the HTTP request body, throw an error
```
marinatedpork: commented out code can be deleted

cmd/lambda/leases/main.go
```diff
@@ -34,74 +61,117 @@ type messageBody struct {
...
+	leaseAddedTopicARN = aws.String(config.GetEnvVar("PROVISION_TOPIC", "DCEDefaultProvisionTopic"))
+	accountDeletedTopicArn = config.GetEnvVar("DECOMMISSION_TOPIC", "DefaultDecomissionTopic")
+	resetQueueURL = config.GetEnvVar("RESET_SQS_URL", "DefaultResetSQSURL")
+	principalBudgetAmount = aws.Float64(config.GetEnvFloatVar("PRINCIPAL_BUDGET_AMOUNT", 1000.00))
+	principalBudgetPeriod = aws.String(config.GetEnvVar("PRINCIPAL_BUDGET_PERIOD", Weekly))
+	maxLeaseBudgetAmount = aws.Float64(config.GetEnvFloatVar("MAX_LEASE_BUDGET_AMOUNT", 1000.00))
+	maxLeasePeriod = aws.Int64(int64(config.GetEnvIntVar("MAX_LEASE_PERIOD", 704800)))
+	defaultLeaseLengthInDays = aws.Int(config.GetEnvIntVar("DEFAULT_LEASE_LENGTH_IN_DAYS", 7))
+
+	leasesRoutes := api.Routes{
```
marinatedpork: Correct me if I am wrong, but with the new routing logic here for the leases API one can no longer query `/leases?principalId=1&status=Active` - right? These query parameters now looks exclusive.

marinatedpork: NVM - The way the router works is that if it doesn’t find an exact match on the query params it falls through to the one without

PR #158, opened 2019-12-09
Move the config service

kddejong:
## Proposed changes

- Move to the config service

## Types of changes

## Checklist


cmd/lambda/accounts/create.go
```diff
@@ -56,100 +56,102 @@ type CreateRequest struct {
...
-
-type principalPolicyInput struct {
-	PrincipalPolicyArn   string
-	PrincipalRoleArn     string
-	PrincipalIAMDenyTags []string
-	AdminRoleArn         string
-	Regions              []string
-}
+//
+//// Validate - Checks if the Account Request has the provided id and adminRoleArn
```
nathanagood: All this commented-out code--would it be better to just remove it? Not sure if this is intentional.

cmd/lambda/accounts/delete.go
```diff
@@ -1,93 +1,97 @@
...
 func DeleteAccount(w http.ResponseWriter, r *http.Request) {
 
 	accountID := mux.Vars(r)["accountId"]
-	deletedAccount, err := Dao.DeleteAccount(accountID)
+	var dao *data.Account
+	if err := Services.Config.GetService(dao); err != nil {
+		ErrorHandler(w, err)
+		return
+	}
+	account, err := account.GetAccountByID(accountID, dao)
```
nathanagood: Usually, when you're injecting an interface for DI it's much more common than not to have the injected arg be the first arg in the method. Consider instead account.GetAccountByID(dao, accountID). I'll also make note on the actual data.Account method.

cmd/lambda/accounts/delete.go
```diff
@@ -1,93 +1,97 @@
...
-	// Handle DB errors
+	err = account.Delete(dao)
 	if err != nil {
 		ErrorHandler(w, err)
 		return
 	}
 
 	// Delete the IAM Principal Role for the account
-	destroyIAMPrincipal(deletedAccount)
+	//destroyIAMPrincipal(deletedAccount)
```
nathanagood: Consider just removing the commented-out code.

pkg/config/servicebuilder.go
```diff
@@ -103,6 +104,12 @@ func (bldr *ServiceBuilder) WithStorageService() *ServiceBuilder {
 	return bldr
 }
 
+// WithDataService tells the builder to add the Data service to the `ConfigurationBuilder`
+func (bldr *ServiceBuilder) WithDataService() *ServiceBuilder {
```
nathanagood: Since this appears to be just creating the data service for account, is this name too generic?

PR #174, opened 2019-12-19
WIP - Feature/account controller redux

kddejong:
## Proposed changes
Lots to digest here.
- Move business logic into a pkg level controller
- Abstract away the data layer and use Go implicit interfaces to work with the data layer
- Simplify the data layer and how queries are built and executed
- Use GO error wrapping and create an error handler to decide exit codes
- Remove redundant Account structs and use one struct for json, dynamodb, and parsing api queries
- Normalization of the config service via Nathan's work

Currently migrated portions of the code
Accounts POST, GET

This that will be deprecated via this work.
- pkg/db
- pkg/api/response/account, lease, usage, and a lot of helpers

Remaining work:
- Create an event engine similar to the data engine.  Let that pkg decide where things go and when they go. 
- Finishing moving the logic for create/delete account once the event engine is created
- Migrate leases/usage

## Types of changes

## Checklist


pkg/model/account.go
```diff
@@ -0,0 +1,43 @@
...
+	LastModifiedOn      *int64                 `json:"lastModifiedOn,omitempty" dynamodbav:"LastModifiedOn"   schema:"lastModifiedOn"`                          // Last Modified Epoch Timestamp
+	CreatedOn           *int64                 `json:"createdOn,omitempty"  dynamodbav:"CreatedOn,omitempty"   schema:"createdOn"`                              // Account CreatedOn
+	AdminRoleArn        *string                `json:"adminRoleArn,omitempty"  dynamodbav:"AdminRoleArn"   schema:"adminRoleArn"`                               // Assumed by the master account, to manage this user account
+	PrincipalRoleArn    *string                `json:"principalRoleArn,omitempty"  dynamodbav:"PrincipalRoleArn,omitempty"   schema:"principalRoleArn"`         // Assumed by principal users
+	PrincipalPolicyHash *string                `json:"principalPolicyHash,omitempty" dynamodbav:"PrincipalPolicyHash,omitempty"   schema:"principalPolicyHash"` // The the hash of the policy version deployed
+	Metadata            map[string]interface{} `json:"metadata,omitempty"  dynamodbav:"Metadata,omitempty"   schema:"metadata"`                                 // Any org specific metadata pertaining to the account
```
eschwartz: @kddejong should this be a pointer type?

eg 
```
Metadata            *map[string]interface{} `...`
```

pkg/model/account.go
```diff
@@ -0,0 +1,43 @@
...
+}
+
+// Accounts is multiple of Account
+type Accounts []Account
```
eschwartz: Maybe cut this (per: @marinatedpork 's comments)?

pkg/data/account.go
```diff
@@ -0,0 +1,172 @@
...
+	ConsistentRead bool   `env:"USE_CONSISTENT_READS" envDefault:"false"`
+}
+
+// Update the Account record in DynamoDB
+func (a *Account) Update(account *model.Account, lastModifiedOn *int64) error {
+
+	var expr expression.Expression
+	var err error
+	var returnValue string
+	// lastModifiedOn is 0 on a create
```
eschwartz: should be:

> // lastModifiedOn is nil on a create

pkg/data/account.go
```diff
@@ -0,0 +1,172 @@
...
+)
+
+// Account - Data Layer Struct
+type Account struct {
+	AwsDynamoDB    dynamodbiface.DynamoDBAPI
+	TableName      string `env:"ACCOUNT_DB"`
+	ConsistentRead bool   `env:"USE_CONSISTENT_READS" envDefault:"false"`
+}
+
+// Update the Account record in DynamoDB
```
eschwartz: Maybe call out that this creates the record, if it doesn't exist

pkg/data/account.go
```diff
@@ -0,0 +1,172 @@
...
+		},
+	)
+	var awsErr awserr.Error
+	if gErrors.As(err, &awsErr) {
+		if awsErr.Code() == "ConditionalCheckFailedException" {
+			return fmt.Errorf(
+				"unable to update account %v no account exists with "+
+					"LastModifiedOn=\"%v\": %w",
+				account.ID, *account.LastModifiedOn, errors.ErrConflict)
+		}
```
eschwartz: There's an unhandled error here

kddejong: Should return the error here too.

pkg/data/account.go
```diff
@@ -0,0 +1,172 @@
...
+	"github.com/aws/aws-sdk-go/aws/awserr"
+	"github.com/aws/aws-sdk-go/service/dynamodb"
+	"github.com/aws/aws-sdk-go/service/dynamodb/dynamodbattribute"
+	"github.com/aws/aws-sdk-go/service/dynamodb/dynamodbiface"
+	"github.com/aws/aws-sdk-go/service/dynamodb/expression"
+)
+
+// Account - Data Layer Struct
+type Account struct {
+	AwsDynamoDB    dynamodbiface.DynamoDBAPI
```
eschwartz: Can this just be called `DynamoDB` (are there other providers of DynamoDB than AWS?)

pkg/data/account.go
```diff
@@ -0,0 +1,172 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("error getting accounts by status %s: %s: %w", status, err, errors.ErrInternalServer)
+	}
+
+	accounts := &model.Accounts{}
+	err = dynamodbattribute.UnmarshalListOfMaps(res.Items, accounts)
+	return accounts, err
+}
+
+// GetAccounts Get a list of accounts based on Principal ID
```
eschwartz: check this comment

pkg/data/account.go
```diff
@@ -0,0 +1,172 @@
...
+func (a *Account) GetAccounts(q *model.Account) (*model.Accounts, error) {
+	var expr expression.Expression
```
eschwartz: Let's talk about pagination.
Maybe don't need to get that question answered in this PR, but i don't want to lose track of it

pkg/api/helpers.go
```diff
@@ -0,0 +1,24 @@
...
+	"github.com/gorilla/schema"
+)
+
+// GetStructFromQuery converts r query into a struct
+func GetStructFromQuery(i interface{}, v url.Values) error {
+	var decoder = schema.NewDecoder()
+
+	err := decoder.Decode(i, v)
+	if err != nil {
+		log.Printf("%+v", err)
```
kddejong: Drop log.PrintF

pkg/accountmanager/manager_test.go
```diff
@@ -0,0 +1,12 @@
+package accountmanager
+
+import (
+	"testing"
+)
+
+func TestAssumeRole(t *testing.T) {
+
+	t.Run("should be able to assume role", func(t *testing.T) {
+
```
eschwartz: missing something here?

pkg/accountmanager/iamrole.go
```diff
@@ -0,0 +1,124 @@
...
+func (am *AccountManager) createRole() error {
+	_ = iam.New(am.awsSession, am.awsConfig)
+
+	return nil
+}
+
+// CreateRoleWithPolicy - Create a Role, and attach a policy to it
+func (am *AccountManager) CreateRoleWithPolicy(input *CreateRoleWithPolicyInput) (*CreateRoleWithPolicyOutput, error) {
+	return nil, nil
+	// createRoleRes, err := am.IAM.CreateRole(&iam.CreateRoleInput{
```
eschwartz: ??

pkg/accountmanager/iamrole.go
```diff
@@ -0,0 +1,124 @@
+package accountmanager
```
eschwartz: what about `childaccountmanager`?
It's longer, but I got a little confused, thinking this package was for managing `Account` struct objects

pkg/account/main.go
```diff
@@ -0,0 +1,46 @@
...
+	"github.com/Optum/dce/pkg/common"
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/session"
+	"github.com/aws/aws-sdk-go/service/dynamodb"
+	"github.com/aws/aws-sdk-go/service/dynamodb/dynamodbiface"
+	"github.com/aws/aws-sdk-go/service/sts"
+)
+
+var (
+	awsSession     *session.Session
```
kddejong: Can this be deleted?

pkg/accountmanager/manager.go
```diff
@@ -0,0 +1,74 @@
...
+	if err != nil {
+		return fmt.Errorf("Error creating Principal Policy Arn: %s: %w", err, errors.ErrInternalServer)
+	}
+
+	return nil
+}
+
+// New creates a new account manager struct
+func New(PrincipalRoleName string, PrincipalPolicyName string) (*AccountManager, error) {
+	new := &AccountManager{
```
eschwartz: FWIW, my IDE doesn't like this
image

cmd/lambda/accounts/main.go
```diff
@@ -185,3 +245,44 @@ func newAWSSession() *session.Session {
...
+		code = "RequestValidationError"
+		message = err.Error()
+	} else if gErrors.Is(err, errors.ErrConflict) {
+		status = http.StatusConflict
+		code = "Conflict"
+		message = err.Error()
+	} else {
+		status = http.StatusInternalServerError
+		code = "ServerError"
+		message = err.Error()
```
kddejong: We don't want to dump the Error fully in this scenario.  Could have sensitive data.

AmudaPalani: I like the error dump. What sensitive data are we concerned about?

pkg/model/account.go
kddejong: I'm not sure on that `Map types are reference types, like pointers or slices...`

https://blog.golang.org/go-maps-in-action

cmd/lambda/accounts/main.go
kddejong: This could have been any error bubbled up.  So I think the concern is database connection information, account information, etc.

pkg/data/account.go
kddejong: how isn't it handled?  I might be missing something here.

pkg/data/account.go
kddejong: I have an idea for that too.  I'm not sure it can be handled as cleanly but I think there is a solution that will work and put into a similar model as everything else.

kddejong:
Closing in replacement of #195

Issue #143, opened 2019-11-25
Error requesting SES email identity verification: Invalid email


eschwartz:
`terraform apply` is failing with:

```
Error: Error requesting SES email identity verification: InvalidParameterValue: Invalid email address<STUB>.
        status code: 400, request id: 1265ca55-d539-4571-a909-b77400a19dfc

  on email_identity.tf line 2, in resource "aws_ses_email_identity" "from_email_address":
   2: resource "aws_ses_email_identity" "from_email_address” {
```

...when using default tfvar values.

PR #175, opened 2019-12-19
Update changelog for v0.24.0

eschwartz:
Preparing for release of v0.24.0

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,15 +1,12 @@
 
-## vNext
+## v0.24.0
 
 - Update Status Lambda - budget_check: Terminate lease if spend > Principal budget amount
 - Support `metadata` parameter in `/accounts` API endpoints
 - Add `PUT /accounts/:id` endpoint
 - Fixed bug where child account's DCEPrincipal role trusted itself rather than the master account
-- Add GetUsageByPrincipal
-- Fix default `budget_notification_from_email` TF var (See #143)
-- Replace GetUsageByDateRange with GetUsageByPrincipal method in POST lease validation
-- Add GetUsage with filter options: accountId, principalId, startDate, limit
 - Add `GET /auth` and `GET /auth/{file+}` endpoints for retrieving credentials web page
+- Support query params for `GET /usage` endpoints
 
 ## v0.23.0
 


CHANGELOG.md
```diff
@@ -1,13 +1,11 @@
...
+## v0.24.0
 
 - Update Status Lambda - budget_check: Terminate lease if spend > Principal budget amount
+- Add a new `Orphaned` status to accounts (#93)
 - Support `metadata` parameter in `/accounts` API endpoints
 - Add `PUT /accounts/:id` endpoint
 - Fixed bug where child account's DCEPrincipal role trusted itself rather than the master account
-- Add GetUsageByPrincipal
-- Fix default `budget_notification_from_email` TF var (See #143)
-- Replace GetUsageByDateRange with GetUsageByPrincipal method in POST lease validation
```
eschwartz: > - Add GetUsageByPrincipal
> - Replace GetUsageByDateRange with GetUsageByPrincipal method in POST lease validation

Thinking these aren't user facing changes, so we don't need them in our changelog

> - Fix default `budget_notification_from_email` TF var (See #143)

This was actually a bug that was only introduced since v0.23.0, so there's no actual change here (ie, it's just "committed a bug, and then fixed the bug")

eschwartz:
I ran through our git history, to see if we missed anything.
The only thing I wasn't sure about were the changes in #157 
@nathanagood -- do we need to update anything in the changelog for that PR?

CHANGELOG.md
```diff
@@ -1,14 +1,11 @@
 
-## vNext
+## v0.24.0
 
 - Update Status Lambda - budget_check: Terminate lease if spend > Principal budget amount
+- Add a new `Orphaned` status to accounts (#93)
```
kddejong: We hadn't called it out.  While we created the status there is no current process to put an account in that state.

eschwartz: got it.

CHANGELOG.md
AmudaPalani: Do we want Users to know that GET usage can be done with various filters?

eschwartz:
> Do we want Users to know that GET usage can be done with various filters?

@AmudaPalani Yes, if that changed that would be good to have in the changelog. I just didn't think it was worth mentioning that specific method names changed.
What would you like to see it say in the changelog about API changes?

eschwartz:
thanks @marinatedpork

PR #181, opened 2019-12-23
New dceerr package for going forward error management

kddejong:
## Proposed changes

- New dceerr package for managing all errors going forward
- Allow for dynamic error printing in the API layer

Error wrapping still needs to be solved.  The database simplification work should complete this errors package.
api/response should eventually go away.  The models aren't needed and the response code will move to the api package.

## Types of changes

## Checklist

### DIFF
--- a/cmd/lambda/accounts/get.go
+++ b/cmd/lambda/accounts/get.go
@@ -8,8 +8,10 @@ import (
 
 	"github.com/gorilla/mux"
 
+	"github.com/Optum/dce/pkg/api"
 	"github.com/Optum/dce/pkg/api/response"
 	"github.com/Optum/dce/pkg/db"
+	"github.com/Optum/dce/pkg/dceerr"
 )
 
 // GetAllAccounts - Returns all the accounts.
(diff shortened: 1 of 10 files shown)


cmd/lambda/accounts/get.go
```diff
@@ -20,7 +22,8 @@ func GetAllAccounts(w http.ResponseWriter, r *http.Request) {
 	if err != nil {
 		errorMessage := fmt.Sprintf("Failed to query database: %s", err)
 		log.Print(errorMessage)
-		response.WriteServerErrorWithResponse(w, errorMessage)
+		api.WriteAPIErrorResponse(w, dceerr.NewInternalServer(errorMessage, nil))
```
kddejong: This may not feel that different but when we move more logic away from this layer we can just raise errors and have them given to the customer appropriately.   This is a middle step showing how it will work but this code will get more simple with the next set of work.

kddejong:
failing till #182 is in

kddejong:
Failing now because of #184

PR #186, opened 2019-12-30
New dceerr package for going forward error management

kddejong:
## Proposed changes

- New dceerr package for managing all errors going forward
- Allow for dynamic error printing in the API layer

Error wrapping still needs to be solved.  The database simplification work should complete this errors package.
api/response should eventually go away.  The models aren't needed and the response code will move to the api package.

## Types of changes

## Checklist

### DIFF
--- a/cmd/lambda/accounts/get.go
+++ b/cmd/lambda/accounts/get.go
@@ -8,8 +8,10 @@ import (
 
 	"github.com/gorilla/mux"
 
+	"github.com/Optum/dce/pkg/api"
 	"github.com/Optum/dce/pkg/api/response"
 	"github.com/Optum/dce/pkg/db"
+	"github.com/Optum/dce/pkg/errors"
 )
 
 // GetAllAccounts - Returns all the accounts.
@@ -20,7 +22,8 @@ func GetAllAccounts(w http.ResponseWriter, r *http.Request) {
 	if err != nil {
 		errorMessage := fmt.Sprintf("Failed to query database: %s", err)
 		log.Print(errorMessage)
-		response.WriteServerErrorWithResponse(w, errorMessage)
+		api.WriteAPIErrorResponse(w, errors.NewInternalServer(errorMessage, nil))
+		return
 	}
 
 	// Serialize them for the JSON response.
@@ -43,12 +46,12 @@ func GetAccountByID(w http.ResponseWriter, r *http.Request) {
 	if err != nil {
 		errorMessage := fmt.Sprintf("Failed List on Account Lease %s", accountID)
 		log.Print(errorMessage)
-		response.WriteServerErrorWithResponse(w, errorMessage)
+		api.WriteAPIErrorResponse(w, errors.NewInternalServer(errorMessage, nil))
 		return
 	}
 
 	if account == nil {
-		response.WriteNotFoundError(w)
+		api.WriteAPIErrorResponse(w, errors.NewNotFound("account", accountID))
 		return
 	}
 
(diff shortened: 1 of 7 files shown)


eschwartz:
@kddejong could we keep our branch names all lowercase? 
I wouldn't care, except it's messing up my git CLI  client (can't autocomplete, creates orphaned refs, etc). I think @nathanagood was having similar issues.

kddejong:
> @kddejong could we keep our branch names all lowercase?
> I wouldn't care, except it's messing up my git CLI client (can't autocomplete, creates orphaned refs, etc). I think @nathanagood was having similar issues.

I'll do that going forward.

pkg/dceerr/error.go
```diff
@@ -0,0 +1,208 @@
+package dceerr
```
eschwartz: it feels a little funny having the product name inside the pkg name.
Maybe just some lingering bad feeling around the "Redbox" refactor 😉 

Is the idea to avoid naming conflicts with the stdlib error package?

pkg/dceerr/error.go
```diff
@@ -0,0 +1,208 @@
...
+
+import (
+	"fmt"
+	"github.com/pkg/errors"
+	"io"
+	"net/http"
+)
+
+const (
```
eschwartz: A comment would nice here on what these are about (looks like this are used in API error response `Code` JSON)

pkg/dceerr/error.go
```diff
@@ -0,0 +1,208 @@
...
+)
+
+type detailError struct {
+	Message string `json:"message"`
+	Code    string `json:"code"`
+}
+
+// StatusError is the custom error type we are using.
+// Should satisfy errors interface
+type StatusError struct {
```
eschwartz: Just a naming thought -- `Status` error is actually a model of an JSON API response. 
But the type doesn't actually include information about the HTTP _status_ code.
Maybe `APIError` would be more suitable?

cmd/lambda/accounts/get.go
```diff
@@ -20,7 +22,8 @@ func GetAllAccounts(w http.ResponseWriter, r *http.Request) {
 	if err != nil {
 		errorMessage := fmt.Sprintf("Failed to query database: %s", err)
 		log.Print(errorMessage)
-		response.WriteServerErrorWithResponse(w, errorMessage)
+		api.WriteAPIErrorResponse(w, dceerr.NewInternalServer(errorMessage, nil))
```
eschwartz: I like this model better, that we pass an error object to a response writer.

pkg/dceerr/error.go
eschwartz: > But the type doesn't actually include information about the HTTP status code.

Oh, I'm wrong. The `apiCode` is the http status code.
I just got confused because our response object also has a `Code` property.

What do you think of something like:
```
type APIError struct {
	httpCode int
	cause   error
	Details detailError `json:"error"`
	stack   *stack
}

Comment (kddejong) [reply pkg/dceerr/error.go]: I don't mind that change.

File: pkg/dceerr/error.go (hunk 1 above)
Comment (kddejong) [reply pkg/dceerr/error.go]: good question.  I could switch to `errors` but it conflicts with the global `errors` package.   I tried not to do that.  I agree I don't like the dce portion of it.  I'm game for changing it.

Comment (AmudaPalani) [reply pkg/dceerr/error.go]: I agree. can we remove dce from package name

File: pkg/dceerr/error_test.go [hunk 5]
```diff
@@ -0,0 +1,181 @@
...
+
+var errOriginal = errors.New("original error")
+var errInternalServer = NewInternalServer("error", errOriginal)
+
+func TestNew(t *testing.T) {
+
+	tests := []struct {
+		name               string
+		err                *StatusError
+		expectedMessage    string
```
AmudaPalani: Can we make this into a nested struct? Call it expectedStatusError?

pkg/dceerr/error.go
eschwartz: sure. I don't really have a better name, though, if we don't want to use  `errors`. so it's ok with me

PR #187, opened 2019-12-30
Add eventer service

kddejong:
## Proposed changes

- Create a service to dispense DCE events 

## Types of changes

## Checklist


AmudaPalani:
Hi Kevin,

Can you please clarify what's driving this change to create a service to dispense DCE events? Is this replacing SNS?

kddejong:
@AmudaPalani it isn't to replace SNS or SQS but allow a common framework to publish an event when they occur.  So on CreateAccount event the eventer package will publish to SNS and SQS as configured.  So we can switch to anything like Event Bus in one location and have the events created in a similar manner.

pkg/event/sns.go
```diff
@@ -0,0 +1,59 @@
...
+
+// Publish an event to the topic
+func (s *SnsEvent) Publish(i interface{}) error {
+	bodyJSON, err := json.Marshal(i)
+	if err != nil {
+		return errors.New("Unable to marshal response")
+	}
+
+	// Wrap the body in a SNS message object
+	message, err := json.Marshal(struct {
```
eschwartz: this could be less verbose, just as:

```
message, err := json.Marshal(map[string]string{
		"default": string(bodyJSON),
		"Body": string(bodyJSON)
	})

File: pkg/event/event.go [hunk 2]
```diff
@@ -0,0 +1,130 @@
...
+	SnsClient              snsiface.SNSAPI
+	SqsClient              sqsiface.SQSAPI
+	AccountCreatedTopicArn string `env:"ACCOUNT_CREATED_TOPIC_ARN" defaultEnv:"DefaultAccountCreatedTopicArn"`
+	AccountDeletedTopicArn string `env:"ACCOUNT_DELETED_TOPIC_ARN" defaultEnv:"DefaultAccountDeletedTopicArn"`
+	AccountResetQueueURL   string `env:"RESET_SQS_URL" defaultEnv:"DefaultResetSQSUrl"`
+	LeaseAddedTopicArn     string `env:"LEASE_ADDED_TOPIC" defaultEnv:"DefaultLeaseAddedTopicArn"`
+}
+
+// Eventer is the public interface for publishing events
+type Eventer struct {
```
eschwartz: Maybe another name? Just thinking that `er` suffixed objects are usually interfaces, by convention

eschwartz: `EventHub`?

eschwartz: or just `Events`?

eschwartz:
> having a bit of trouble wrapping 

I think what's sitting a little odd with me is that all our of DCE-specific event types are baked right into a package that is otherwise dealing with generic eventing logic. So for example, if I was trying contribute code to publish an `account-updated` event, I'm not sure I'd think to look in the `NewEventer` function for configuring that.

Also, I'm not sure about how we're kind tacking on `resetAccount` events into our `accountDelete` and `accountCreate` events. I see events as providing integration points where when things _have happened_ in the system. The reset SQS message, on the other hand, is an imperative request -- it's something we want to happen, as a business requirement. So if I go into the `accounts/delete.go` controller, I'd want to see some code that triggers an account reset. Having it in the `Eventer` constructor feels side-effect-y

pkg/event/sns.go
kddejong: Thanks updated.

pkg/event/event.go
kddejong: Switching this to Hub.  EventHub replicates Event from the package.

PR #189, opened 2019-12-31
Update TF and tflint versions

kddejong:
## Proposed changes

- Update Terraform and TFLint Versions

## Types of changes

## Checklist

### DIFF
--- a/modules/alarms_sns.tf
+++ b/modules/alarms_sns.tf
@@ -25,7 +25,7 @@ resource "aws_sns_topic_policy" "alarms_policy" {
           "SNS:DeleteTopic",
           "SNS:AddPermission"
         ],
-        "Resource": "aws_sns_topic.alarms_topic.arn",
+        "Resource": "${aws_sns_topic.alarms_topic.arn}",
         "Condition": {
           "StringEquals": {
             "AWS:SourceOwner": ["${data.aws_caller_identity.current.account_id}"]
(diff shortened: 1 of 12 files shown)


modules/alarms_sns.tf
```diff
@@ -25,7 +25,7 @@ resource "aws_sns_topic_policy" "alarms_policy" {
           "SNS:DeleteTopic",
           "SNS:AddPermission"
         ],
-        "Resource": "aws_sns_topic.alarms_topic.arn",
+        "Resource": "${aws_sns_topic.alarms_topic.arn}",
```
nathanagood: It looks like everywhere else you've _removed_ the `${..}` notation... why adding them here?

kddejong: Its because inside a text blob. ```policy = <<POLICY```.  I believe this is actually just a fix.

nathanagood: Awesome. I'll approve, LGTM

kddejong:
@eschwartz the terraform plan from master to this version is clean.  The fix will be detected for the SNS policy but other than that there are no other changes.

pipelines/pipeline.yml
```diff
@@ -44,9 +44,15 @@ stages:
             displayName: "Set up Golang CI Tools"
 
           # Unit Tests
-          - script: |
-              make test
+          - task: AWSShellScript@1
             displayName: "Unit Tests"
+            inputs:
+              awsCredentials: "AWS_Login"
```
eschwartz: @kddejong what's the story here?
I think we actually need to revert this -- I'll message you in Flowdock with details

eschwartz: band-aid fix put it. We'll have to revisit this, if we start getting some community PRs

PR #195, opened 2020-01-03
account controller redux

kddejong:
## Proposed changes

- Switch to account service
- Add new data service and make working with the data layer easier
- Add more functions to errors to make working with the dce errors package easier

Next steps include rewriting the accounts get and list controller functions

## Types of changes

## Checklist


pkg/account/account.go
```diff
@@ -0,0 +1,339 @@
...
+}
+
+// Manager manages all the actions against an account
+type Manager interface {
+	Setup(adminRole string) error
+}
+
+// Account is a type corresponding to a Account table record
+type Account struct {
+	writer WriterDeleter
```
kddejong: added this to the Account struct.  To @marinatedpork feedback I was passing this into a lot of functions and realized they are all the same.  The only thing that can be done to an account after this struct is instantiated is write/delete so just making it part of the struct instead of a parameter to the functions.  Notice this doesn't include the read commands.

pkg/account/account.go
```diff
@@ -0,0 +1,339 @@
...
+	}
+
+	account := &(*accounts)[0]
+	account.writer = wd
+
+	return account, err
+}
+
+// CreateAccount creates a new account
+func CreateAccount(ID string, AdminRoleArn string, Metadata map[string]interface{}, r Reader, wd WriterDeleter) (*Account, error) {
```
kddejong: This function isn't used currently.  Only here to stub it out as it will be used at some point but more work is needed.  Eventer service needs to get put in and the account manager service needs to get put in.

pkg/data/accounts.go
```diff
@@ -0,0 +1,104 @@
...
+	"github.com/Optum/dce/pkg/errors"
+	"github.com/Optum/dce/pkg/model"
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/service/dynamodb"
+	"github.com/aws/aws-sdk-go/service/dynamodb/dynamodbattribute"
+	"github.com/aws/aws-sdk-go/service/dynamodb/expression"
+)
+
+// queryAccounts for doing a query against dynamodb
+func (a *Account) queryAccounts(q *model.Account, keyName string, index string) (*model.Accounts, error) {
```
kddejong: Might move this to a helper function as we add leases and usage.  There isn't much here that is particular to accounts, leases, usage, etc.

pkg/data/accounts.go
```diff
@@ -0,0 +1,104 @@
...
+	accounts := &model.Accounts{}
+	err = dynamodbattribute.UnmarshalListOfMaps(res.Items, accounts)
+	if err != nil {
+		return nil, errors.NewInternalServer("failed unmarshaling of accounts", err)
+	}
+	return accounts, err
+}
+
+// GetAccounts Get a list of accounts
+func (a *Account) GetAccounts(q *model.Account) (*model.Accounts, error) {
```
kddejong: Generic GetAccounts.  This has the logic how to handle querying an account.  What indexes to use when certain query logic is presentable.

I think we should hard error when someone tries to query on AccountID as they should be doing /accounts/accountID/

pkg/data/helpers.go
```diff
@@ -0,0 +1,39 @@
+package data
+
+import (
+	"reflect"
+	"strings"
+
+	"github.com/aws/aws-sdk-go/service/dynamodb/expression"
+)
+
+func getFiltersFromStruct(i interface{}, keyName *string) (*expression.KeyConditionBuilder, *expression.ConditionBuilder) {
```
kddejong: Build filters and key condition (filters).  Using the keyName to determine if a column is part of an index and switch it from a filter to a key condition.

pkg/errors/wrap.go
```diff
@@ -0,0 +1,39 @@
+package errors
```
kddejong: added this to remove the need for the "errors" package.  Should be able to do everything we need without that package anymore.

pkg/model/account.go
```diff
@@ -0,0 +1,43 @@
...
+
+// Accounts is multiple of Account
+type Accounts []Account
+
+// AccountStatus is an account status type
+type AccountStatus string
+
+const (
+	// None status
+	None AccountStatus = "None"
```
kddejong: Wondering if we should have these as AccountNone, AccountReady

pkg/account/account.go
```diff
@@ -0,0 +1,339 @@
...
+	return a.data.Metadata
+}
+
+func (a *Account) updateStatus(nextStatus model.AccountStatus) {
+	*a.data.Status = nextStatus
+}
+
+// UpdateStatus updates account status for a given accountID and
+// returns the updated record on success
+func (a *Account) UpdateStatus(nextStatus model.AccountStatus) error {
```
kddejong: Hoping to get rid of this for a more generic update function

pkg/account/validate.go
```diff
@@ -0,0 +1,57 @@
...
+)
+
+// We don't use the internal errors package here because validation will rewrite it anyways
+// Just spit out errors and turn them into validation errors inside the appropriate functions
+
+var accountIDRule = []validation.Rule{
+	validation.Match(regexp.MustCompile("^[0-9]{12}$")).Error("must be a string with 12 digits"),
+}
+
+func isNil(value interface{}) error {
```
kddejong: May want to move some of these to a generic package so they can be shared with leases, usage or this account package needs to be generalized into a new package name.

pkg/data/account.go
```diff
@@ -0,0 +1,132 @@
...
+
+// Account - Data Layer Struct
+type Account struct {
+	DynamoDB       dynamodbiface.DynamoDBAPI
+	TableName      string `env:"ACCOUNT_DB"`
+	ConsistentRead bool   `env:"USE_CONSISTENT_READS" envDefault:"false"`
+}
+
+// WriteAccount the Account record in DynamoDB
+func (a *Account) WriteAccount(account *model.Account, lastModifiedOn *int64) error {
```
kddejong: lastModifiedOn here is the previous lastModifiedOn.  not sure that is clear enough.

eschwartz:
Taking a 👀 at this now

pkg/account/account.go
```diff
@@ -0,0 +1,318 @@
...
+	return nil
+}
+
+// Update the Account record in DynamoDB
+func (a *Account) Update(d model.Account, am Manager) error {
+	err := validation.ValidateStruct(&d,
+		// ID has to be empty
+		validation.Field(&d.ID, validation.NilOrNotEmpty, validation.In(*a.data.ID)),
+		validation.Field(&d.AdminRoleArn, validation.By(isNilOrUsableAdminRole(am))),
+		//validation.Field(&d.AdminRoleArn, validation.By(isNil)),
```
eschwartz: can remove commented out code

pkg/account/account.go
```diff
@@ -0,0 +1,318 @@
...
+	if !errors.Is(err, &errors.StatusError{}) {
+		if err != nil {
+			return nil, fmt.Errorf("Failed to add account %s to pool: %w", ID, err)
+		}
+		if existingAccount.data.ID != nil {
+			return nil, errors.NewConflict("account", ID, nil)
+		}
+	}
+
+	// Verify that we can assume role in the account,
```
eschwartz: can remove commented out code

pkg/account/account.go
```diff
@@ -0,0 +1,237 @@
...
+}
+
+// MarshalJSON Marshals the data inside the account
+func (a *Account) MarshalJSON() ([]byte, error) {
+	return json.Marshal(a.data)
+}
+
+// GetReadyAccount returns an available account record with a
+// corresponding status of 'Ready'
+func GetReadyAccount(d Reader, wd WriterDeleter) (*Account, error) {
```
eschwartz: Could we pass in a single `ReaderWriterDeleter` object here?

pkg/account/account.go
```diff
@@ -0,0 +1,237 @@
...
+	accounts, err := GetAccounts(
+		&model.Account{
+			Status: model.AccountStatusReady.AccountStatusPtr(),
+		}, d)
+	if err != nil {
+		return nil, err
+	}
+	if len(accounts.data) < 1 {
+		return nil, errors.NewNotFound("account", "ready")
```
eschwartz: Thoughts on returning an error here, vs returning `nil, nil`?

I would probably lean towards returning nil, and letting the controller decide what to do with that.
But it probably doesn't matter too much what we do, if we're consistent

pkg/data/account.go
```diff
@@ -0,0 +1,146 @@
...
+)
+
+// Account - Data Layer Struct
+type Account struct {
+	DynamoDB       dynamodbiface.DynamoDBAPI
+	TableName      string `env:"ACCOUNT_DB"`
+	ConsistentRead bool   `env:"USE_CONSISTENT_READS" envDefault:"false"`
+}
+
+// WriteAccount the Account record in DynamoDB
```
eschwartz: This could probably use some expanded comments, 
since some of the behavior might not be assumed, and it's core to the system.
eg., that:
- It's an "upsert" operation (creates or updates existing)
- we assume it's a create if `lastModified` is nil

pkg/data/account.go
eschwartz: Maybe make explicit? Like `prevLastModifiedOn`?
Also, could elaborate in method comments

pkg/data/account.go
```diff
@@ -0,0 +1,146 @@
...
+			ReturnValues: aws.String(returnValue),
+		},
+	)
+	var awsErr awserr.Error
+	if gErrors.As(err, &awsErr) {
+		if awsErr.Code() == "ConditionalCheckFailedException" {
+			return errors.NewConflict(
+				"account",
+				*account.ID,
+				fmt.Errorf("unable to update account with LastModifiedOn=%q", strconv.FormatInt(*account.LastModifiedOn, 10)))
```
eschwartz: I like this check here, as it prevents race conditions if we were having many writes at once.
But this message might be confusing for consumers. 
Maybe something like: "unable to update account: accounts has been modified since request was made"

pkg/data/account.go
```diff
@@ -0,0 +1,146 @@
...
+			return errors.NewInternalServer("error building query", err)
+		}
+		returnValue = "NONE"
+	} else {
+		modExpr := expression.Name("LastModifiedOn").AttributeNotExists()
+		expr, err = expression.NewBuilder().WithCondition(modExpr).Build()
+		if err != nil {
+			return errors.NewInternalServer("error building query", err)
+		}
+		returnValue = "NONE"
```
eschwartz: Looks like this is always set to `"NONE"`?

eschwartz: On that note -- I'm thinking a `POST /accounts` or `PUT /accounts` request should return the new account object, so maybe this needs to return `"ALL_NEW"` in order to populate that object.

eschwartz: or, the controller needs to run an extra query, to grab the account info.

pkg/data/accounts.go
```diff
@@ -0,0 +1,96 @@
...
+
+	keyCondition, filters := getFiltersFromStruct(q, &keyName)
+	bldr = expression.NewBuilder().WithKeyCondition(*keyCondition)
+	if filters != nil {
+		bldr = bldr.WithFilter(*filters)
+	}
+
+	expr, err = bldr.Build()
+	if err != nil {
+		return nil, errors.NewInternalServer("unabled to build query", err)
```
eschwartz: typo ("unabled" --> "unable")

pkg/data/accounts.go
```diff
@@ -0,0 +1,96 @@
...
+func (a *Account) scanAccounts(q *model.Account) (*model.Accounts, error) {
+	var expr expression.Expression
+	var err error
+	var res *dynamodb.ScanOutput
+
+	_, filters := getFiltersFromStruct(q, nil)
+	if filters != nil {
+		expr, err = expression.NewBuilder().WithFilter(*filters).Build()
+		if err != nil {
+			return nil, errors.NewInternalServer("unabled to build query", err)
```
eschwartz: typo ("unabled" --> "unable")

pkg/account/accounts.go
```diff
@@ -0,0 +1,34 @@
...
+}
+
+// GetAccounts Get a list of accounts based on Principal ID
+func GetAccounts(q *model.Account, d MultipleReader) (*Accounts, error) {
+	err := validation.ValidateStruct(q,
+		// ID has to be empty
+		validation.Field(&q.ID, validation.NilOrNotEmpty, validation.By(isNil)),
+	)
+	if err != nil {
+		return nil, errors.NewValidation("account", err)
```
eschwartz: What will this validation error look like to the end user?
Thinking that an API response like `"account validation error: ID must not be nil"` would be pretty confusing.
Maybe better to do our own check (here, or in controller), and response like `"'id' is not a supported query parameter"`

pkg/data/accounts.go
```diff
@@ -0,0 +1,96 @@
...
+	if err != nil {
+		return nil, errors.NewInternalServer("failed unmarshaling of accounts", err)
+	}
+	return accounts, err
+}
+
+// GetAccounts Get a list of accounts
+func (a *Account) GetAccounts(q *model.Account) (*model.Accounts, error) {
+
+	if q.Status != nil {
```
eschwartz: 👍 

If we wanted to be really fancy, we could probably include this info in the struct annotation,
and then figure out on the fly if this is a query or a scan (and what index to use)

pkg/data/account_test.go
```diff
@@ -0,0 +1,251 @@
...
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			mockDynamo := awsmocks.DynamoDBAPI{}
+
+			mockDynamo.On("PutItem", mock.MatchedBy(func(input *dynamodb.PutItemInput) bool {
+				if tt.oldLastModifiedOn == nil {
+					return (*input.TableName == "Accounts" &&
```
eschwartz: You can actually throw `assert.Equal()` statements into these mock callbacks, instead of having to write out big conditionals like this. It gives you a nicer messages, then, if any of your conditions are false.

Just don't use `require.Equal()` -- it breaks your test in a gross way.

pkg/account/account.go
kddejong: I was thinking about that but we only keep the wd and it isn't easy to split apart.  The `wd` was a fallout on removing it from being passed into so many functions.

The other option would be to have a Setter type function that could be used to set it when needed.

After discussion I'm going to remove this function.

pkg/data/accounts.go
kddejong: Agreed.

pkg/data/account.go
kddejong: Switched it.

pkg/account/account.go
kddejong: Removed this function but this is a good discussion topic.

pkg/data/account.go
kddejong: Updated.

pkg/data/account.go
kddejong: The current logic works with NONE.  The big debate is around the patch or DynamoDBs UpdateItem vs PutItem.  Right now we use PutItem which means we have to have the complete record before the update.

pkg/account/accounts.go
kddejong: Updated.

pkg/data/account.go
```diff
@@ -24,26 +23,27 @@ type Account struct {
 }
 
 // WriteAccount the Account record in DynamoDB
-func (a *Account) WriteAccount(account *model.Account, lastModifiedOn *int64) error {
+// This is an upsert operation in which the record will either
+// be inserted or updated
+// prevLastModifiedOn parameter is the original lastModifiedOn
+func (a *Account) WriteAccount(account *model.Account, prevLastModifiedOn *int64) error {
```
eschwartz: 👍

pkg/data/account.go
```diff
@@ -67,7 +67,7 @@ func (a *Account) WriteAccount(account *model.Account, lastModifiedOn *int64) er
 			return errors.NewConflict(
 				"account",
 				*account.ID,
-				fmt.Errorf("unable to update account with LastModifiedOn=%q", strconv.FormatInt(*account.LastModifiedOn, 10)))
+				fmt.Errorf("unable to update account: accounts has been modified since request was made"))
```
eschwartz: 👍

eschwartz:
latest changes look good. Just ping me when the build is passing, and I'll give it a 👍

kddejong:
Looks like my terraform state file got messed up. Going to do some more manual cleanup to get it green

pkg/errors/error_test.go
```diff
@@ -137,7 +138,53 @@ func TestNew(t *testing.T) {
...
+			assert.NotNil(
+				t,
+				GetStackTraceForError(tt.err),
+			)
+		})
+	}
+}
+
+func TestNewGeneric(t *testing.T) {
+
```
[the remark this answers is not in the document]
kddejong: Its tested with https://github.com/Optum/dce/pull/195/files#diff-e9ad42192ee041e5f9f80e25270ac1c4R10  

I can create a more specific test but the current coverage of this package is 98.7% and that function is 100%.

pkg/errors/wrap.go
```diff
@@ -0,0 +1,39 @@
...
+		Unwrap() error
+	})
+	if !ok {
+		return nil
+	}
+	return u.Unwrap()
+}
+
+// Is for seeing if an error includes the target error
+func Is(err, target error) bool {
```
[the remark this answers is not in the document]
kddejong: This Is a generic package Is. 

We use this as errors.Is(err, typeoferr).  This will call the included Is function that you have highlighted if it exists for the Error.  Not all errors are required to have Is implemented so this provides that logic.

pkg/model/account.go
```diff
@@ -0,0 +1,49 @@
+package model
+
+// Account - Handles importing and exporting Accounts and non-exported Properties
+type Account struct {
+	ID                  *string                `json:"id,omitempty" dynamodbav:"Id" schema:"id"`                                                                // AWS Account ID
```
AmudaPalani: What's the purpose schema tag? Please clarify.

AmudaPalani: I found answer to my question. Package gorilla/schema fills a struct with form values based on schema tag.

pkg/model/account.go
```diff
@@ -0,0 +1,49 @@
...
+
+// Accounts is multiple of Account
+type Accounts []Account
+
+// AccountStatus is an account status type
+type AccountStatus string
+
+const (
+	// AccountStatusNone status
+	AccountStatusNone AccountStatus = "None"
```
AmudaPalani: Can we remove const that are not used?

AmudaPalani: AccountStatusNone, AccountStatusLeased, AccountStatusOrphaned

pkg/model/account.go
AmudaPalani: https://stackoverflow.com/questions/50536368/does-gorilla-schema-support-the-json-reflection-tag-natively

It looks like we can use the same json tag too?

pkg/errors/wrap_test.go
```diff
@@ -0,0 +1,56 @@
...
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestIs(t *testing.T) {
+
+	tests := []struct {
+		name        string
+		originalErr error
```
AmudaPalani: Instead of calling this as originalErr and isErr, can we rename these as actualErr and expectedErr?

pkg/data/accounts.go
```diff
@@ -0,0 +1,96 @@
...
+		KeyConditionExpression:    expr.KeyCondition(),
+		ConsistentRead:            aws.Bool(a.ConsistentRead),
+		FilterExpression:          expr.Filter(),
+		ExpressionAttributeNames:  expr.Names(),
+		ExpressionAttributeValues: expr.Values(),
+	})
+
+	if err != nil {
+		return nil, errors.NewInternalServer(
+			fmt.Sprintf("failed to get accounts by status %q", *q.Status),
```
AmudaPalani: How do we know it's get account by status is the filter used?

pkg/model/account.go
kddejong: They will be used.  I can remove them now but they will be added back very shortly... once I rewrite cmd/lamba/accounts/update

pkg/data/accounts.go
kddejong: Good catch.  Corrected.

pkg/errors/wrap_test.go
kddejong: left it as originalErr since that error is wrapped.  Both are actual errors.  

switch isErr to expErr

pkg/model/account.go
kddejong: Removing the schema for now.  Good find on using the json tag.  I'll try it out when that part gets implemented or add back the scheme.  

Some of this is leftover of breaking out the large PR into smaller ones.  Not everything is getting used in this state but the next PR will use it.

pkg/model/account.go
kddejong: I'm going to resolve this one for now.  Not a lot of gain here.

PR #206, opened 2020-01-08
DCEify PR accounts

eschwartz:
## Proposed changes

Trying to run our PR DCE deployments inside a DCE child account.

**Driving this change:**
- Our PR account has a ton of orphaned resources, which costs $$, presents security risk, and hits AWS resource limits (causing failed builds)
- Dogfooding -- can we use DCE as we intend our customers to use it? testing out the dce-cli

**Overview of changes**

- Before TF deploy, new step to get access to a DCE child account
- Creates a lease for the PR, with `PrincipalID=github-pr-<pr#>`, or uses existing lease if already created
- "Logs in" to that lease, generating AWS creds for the DCE child account
- Those child creds are used in future TF deploy/functional-test steps.

**Also..**

- Remove access logging from our artifacts bucket. The number of objects this generates made nuking difficult, and there wasn't a compelling reason for us to keep it,.

## Types of changes

## Checklist

### DIFF
--- a/modules/artifacts_bucket.tf
+++ b/modules/artifacts_bucket.tf
@@ -29,28 +29,6 @@ resource "aws_s3_bucket" "artifacts" {
     enabled = true
   }
 
-  # Send S3 access logs for this bucket to itself
-  logging {
-    target_bucket = local.artifact_bucket_name
-    target_prefix = "logs/"
-  }
-
-  // Expire logs after 7 days
-  lifecycle_rule {
-    id      = "logs"
-    prefix  = "logs/"
-    enabled = true
-
-    expiration {
-      days = 7
-    }
-
-    // Expire previous object versions after 1 day
-    noncurrent_version_expiration {
-      days = 1
-    }
-  }
-
   tags = var.global_tags
 }
 
--- a/pipelines/pipeline.yml
+++ b/pipelines/pipeline.yml
@@ -24,12 +24,6 @@ stages:
               addToPath: true
               architecture: "x64"
 
-          # Install AWS ClI tool
-          - task: CmdLine@2
-            inputs:
-              script: "pip install awscli"
-            displayName: "Install awscli"
-
           # Install Terraform
           - task: TerraformInstaller@0
             inputs:
@@ -76,23 +70,17 @@ stages:
       - job: Build
         dependsOn: Test
         steps:
+          # Build go executables
           - script: |
               make build
             displayName: "Build"
+
+          # Publish build artifacts
+          # for deploying to PR environment, and for Github releases
           - publish: $(System.DefaultWorkingDirectory)/bin
             artifact: bin
-          - publish: $(System.DefaultWorkingDirectory)/modules
-            artifact: modules
           - publish: $(System.DefaultWorkingDirectory)/scripts
             artifact: deploy_scripts
-          - publish: $(System.DefaultWorkingDirectory)/tests
-            artifact: tests
-          - publish: $(System.DefaultWorkingDirectory)/pkg
-            artifact: pkg
-          - publish: $(System.DefaultWorkingDirectory)/go.mod
-            artifact: gomod
-          - publish: $(System.DefaultWorkingDirectory)/go.sum
-            artifact: gosum
 
   - stage: Deploy
     dependsOn: TestAndBuild
@@ -104,6 +92,9 @@ stages:
           runOnce:
             deploy:
               steps:
+                # Checkout the git repo code
+                - checkout: self
+
                 # Configure the namespace used by Terraform
                 # May be referenced as an env var (eg. "${NAMESPACE}")
                 # Or as a pipeline variable (eg. "$(namespace)")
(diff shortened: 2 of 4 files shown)


pipelines/pipeline.yml
```diff
@@ -13,89 +13,89 @@ pr:
   - release/*
 
 stages:
-  - stage: TestAndBuild
```
eschwartz: commented out a bunch of the pipeline, just to make iteration faster while I'm working on this.

pipelines/pipeline.yml
```diff
@@ -184,37 +138,160 @@ stages:
...
+                      echo "
+                        api:
+                          # nonprod (0086)
+                          host: $(DCE_NONPROD_HOSTNAME)
+                          basepath: /api
+                        region: us-east-1
+                      " > ./dce.yml
+
+
+                      # Check if we already have an active lease for this PR
```
eschwartz: Half this bash code could deleted if we fixed a couple bugs in the dce-cli 
(outputing to stderr, invalid JSON, unsupported `GET` query params, etc)

pipelines/pipeline.yml
```diff
@@ -76,23 +70,17 @@ stages:
...
+
+          # Publish build artifacts
+          # for deploying to PR environment, and for Github releases
           - publish: $(System.DefaultWorkingDirectory)/bin
             artifact: bin
-          - publish: $(System.DefaultWorkingDirectory)/modules
-            artifact: modules
           - publish: $(System.DefaultWorkingDirectory)/scripts
             artifact: deploy_scripts
-          - publish: $(System.DefaultWorkingDirectory)/tests
```
eschwartz: I mucked around a bunch with how files are shared between jobs.
I added `checkout: true` to all the stages, so everything has access to the main repo code,
and we only need to move generated artifacts around.

Not necessarily better or worse this way, but I was having a rough time of it, 
and it's working now!

pipelines/pipeline.yml
```diff
@@ -123,59 +114,16 @@ stages:
...
-                      backend "s3" {
-                        region = "us-east-1"
-                      }
-                    }
-                    ' > backend.tf
-                  displayName: "Generate backend.tf"
-                  workingDirectory: "$(Pipeline.Workspace)/modules"
-
-                # terraform init
-                - task: TerraformTaskV1@0
```
eschwartz: We can't really use these TerraformTasks anymore, because they require static credentials. We want to use our dynamic creds from our DCE lease.

They didn't add much value anyways -- an unnecessary abstraction around a single CLI command, imo

scripts/create-tf-backend.sh
```diff
@@ -0,0 +1,62 @@
+#!/usr/bin/env bash
```
eschwartz: Because we're using ephemeral accounts in PR envs, we need to recreate our TF backend whenever we get a clean account.

pipelines/pipeline.yml
kddejong: Tried to find that type of property but was having really bad luck.  Nice find.

pipelines/pipeline.yml
```diff
@@ -186,35 +134,157 @@ stages:
...
                     testRunTitle: Functional Tests
                   condition: always()
+
+                # Publish the dce-cli executable, so we can use
+                # it in our Cleanup stage
+                - publish: $(System.DefaultWorkingDirectory)/dce
+                  artifact: dce-cli
+                - publish: $(System.DefaultWorkingDirectory)/dce.yml
+                  artifact: dce-yml
+                - publish: $(System.DefaultWorkingDirectory)/lease_id.txt
```
kddejong: Can't we just requery this?

pipelines/pipeline.yml
eschwartz: yeah, and it's weird because _some_ stages do it automatically, and some don't.

pipelines/pipeline.yml
```diff
@@ -186,35 +134,157 @@ stages:
...
+                  displayName: "Download pipeline dependencies"
+
+                # Lease a DCE account, to use for deploying our PR environment
+                # (deploy DCE in DCE)
                 - task: AWSShellScript@1
-                  displayName: "Functional Tests"
+                  displayName: "Lease DCE Account"
                   inputs:
-                    awsCredentials: "AWS_Login"
+                    awsCredentials: "AWS_DCE_API_NONPROD"
```
kddejong: I'm guessing you added this credential to the pipeline already?

scripts/create-tf-backend.sh
```diff
@@ -0,0 +1,62 @@
...
+  # Create the S3 bucket
+  echo "Creating S3 bucket s3://${bucket_name} to use as terraform state backend... "
+  aws s3api create-bucket \
+    --bucket ${bucket_name} \
+    --acl log-delivery-write > /dev/null
+  echo "done."
+
+  # Operations on the bucket seem to fail if immediately after creating
+  # the bucket.
+  sleep 20
```
kddejong: ouch :(

modules/artifacts_bucket.tf
```diff
@@ -29,28 +29,6 @@ resource "aws_s3_bucket" "artifacts" {
     enabled = true
   }
 
-  # Send S3 access logs for this bucket to itself
-  logging {
```
kddejong: Thank you!

pipelines/pipeline.yml
```diff
@@ -225,70 +295,68 @@ stages:
...
-                # terraform destroy
-                - task: TerraformTaskV1@0
-                  displayName: "terraform destroy"
-                  inputs:
-                    provider: "aws"
-                    command: "destroy"
-                    workingDirectory: "$(Pipeline.Workspace)/modules"
-                    commandOptions: '-var="namespace=$(namespace)" -var="budget_notification_from_email=$(NOTIFY_EMAIL)"'
-                    environmentServiceNameAWS: "AWS_Terraform_Connection"
+                - bash: |
```
kddejong: Should we be ending the lease?

modules/artifacts_bucket.tf
joshmarsh: 💯

scripts/create-tf-backend.sh
eschwartz: `sleep` is the unsung hero of software 😜

pipelines/pipeline.yml
eschwartz: yeah, we could. There are just some things in the current version of dce-cli that make it difficult to lookup a specific lease (printing to stderr, can't query by status and principal..), so I was trying to avoid it for now.

PR #210, opened 2020-01-09
Convert accounts GET controller to new data model

kddejong:
## Proposed changes

- Use new data model for accounts GET controller
- Build out ServiceBuilder for additionally needed items

## Types of changes

## Checklist


pkg/errors/error.go
```diff
@@ -33,21 +33,21 @@ type StatusError struct {
 	stack    *stack
 }
 
-func (e *StatusError) Error() string { return e.Details.Message }
```
kddejong: Makes a lot more sense here to not use Pointer receivers for these functions.

pkg/data/dataiface/account.go
```diff
@@ -0,0 +1,20 @@
+//
```
kddejong: Common dataiface to make working with the servicebuilder easier.

cmd/lambda/accounts/main.go
```diff
@@ -123,6 +138,36 @@ func init() {
...
+
+	_, err = svcBldr.
+		// AWS services...
+		WithDynamoDB().
+		WithSTS().
+		WithS3().
+		WithSNS().
+		WithSQS().
+		// DCE services...
+		WithStorageService().
```
joshmarsh: Seems redundant to be building both `WithS3()` and `WithStorageService()`

kddejong: Yea, it can be considered redundant.  But the DCE services will utilize the AWS services.  StorageService has the ability to read go template files from S3 but it uses the S3 service.

PR #211, opened 2020-01-10
add new data model to account update controller

kddejong:
## Proposed changes

- Migrate accounts update controller to use pkg/account and pkg/data
- Add account manager pkg to better support interface/API design

## Types of changes

## Checklist


pkg/account/account.go
```diff
@@ -185,27 +185,28 @@ func (a *Account) Delete() error {
...
-	newAccount.data = *data
 
-	return &newAccount, err
+	return New(*data, wd), err
 }
 
 // New returns an account from ID
-func New(wd WriterDeleter, data model.Account) *Account {
-	now := time.Now().Unix()
+func New(data model.Account, wd WriterDeleter) *Account {
```
kddejong: For consistency.  I've been doing interfaces at the end.

tests/acceptance/api_test.go
```diff
@@ -1092,8 +1092,6 @@ func TestApi(t *testing.T) {
 					"metadata": map[string]interface{}{
 						"foo": "bar",
 					},
-					// this shouldn't actually persist
-					"accountStatus": "hippos",
```
kddejong: accountStatus isn't settable via update.  Previously this would just skip and stlil write the rest of the data.  Going forward this will cause a validation error.

tests/acceptance/api_test.go
```diff
@@ -1148,7 +1141,7 @@ func TestApi(t *testing.T) {
 			// with invalid adminRoleArn
 			res := apiRequest(t, &apiRequestInput{
 				method: "PUT",
-				url:    apiURL + "/accounts/not-an-account-id",
+				url:    apiURL + "/accounts/123456789012",
```
kddejong: accountIDs will have to be 12 digit #s going forward.  This will fail validation before it will even try to get an account.

pkg/accountmanager/manager.go
```diff
@@ -0,0 +1,74 @@
...
+	sts                 stsiface.STSAPI
+	adminRoleArn        arn.ARN
+	principalRoleName   string `env:"PRINCIPAL_ROLE_NAME"`
+	principalPolicyName string `env:"PRINCIPAL_POLICY_NAME"`
+	principalRoleArn    arn.ARN
+	principalPolicyArn  arn.ARN
+}
+
+// Setup creates a new session manager struct
+func (am *AccountManager) Setup(adminRoleArn string) error {
```
kddejong: The idea here is a single struct that can act against the child accounts.  It uses awsConfig when setting up clients.  This way credentials are obtained once and automatically rotated without additional configuration.  This should simplify account access and credentials, role gathering.

cmd/lambda/accounts/update.go
```diff
@@ -2,116 +2,53 @@ package main
...
-	Metadata            *map[string]interface{} `json:"metadata"`
-}
-
+// UpdateAccountByID updates an accounts information based on ID
 func UpdateAccountByID(w http.ResponseWriter, r *http.Request) {
 	accountID := mux.Vars(r)["accountId"]
 
 	// Deserialize the request JSON as an request object
-	var request *updateAccountRequest
+	d := model.Account{}
```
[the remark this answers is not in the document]
kddejong: Will do.

pkg/accountmanager/accountmanageriface/manager.go
```diff
@@ -0,0 +1,9 @@
+//
```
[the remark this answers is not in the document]
kddejong: This was done by an auto generator

eschwartz:
@kddejong taking a :eyes: at this now....

pkg/accountmanager/manager.go
```diff
@@ -0,0 +1,74 @@
...
+			am.adminRoleArn.AccountID, am.principalPolicyName))
+	if err != nil {
+		return errors.NewInternalServer("error creating principal policy arn", err)
+	}
+
+	return nil
+}
+
+// NewInput holds the configuration for a new AccountManager
+type NewInput struct {
```
eschwartz: this is an interesting pattern, to define the env vars right in the constructor input.

cmd/lambda/accounts/update_test.go
```diff
@@ -1,325 +1,129 @@
...
+	amMocks "github.com/Optum/dce/pkg/accountmanager/accountmanageriface/mocks"
+	"github.com/Optum/dce/pkg/config"
+	dataMocks "github.com/Optum/dce/pkg/data/dataiface/mocks"
+	"github.com/Optum/dce/pkg/model"
+	"github.com/gorilla/mux"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/mock"
 )
 
-func TestUpdateAccountByID(t *testing.T) {
```
eschwartz: @kddejong -- what's the story here? There's behavior captured in these tests that I'm not seeing in the new test coverage, and it's not clear if the refactored controller implements the behavior.

I'd like to avoid breaking changes here if we can...

cmd/lambda/accounts/update.go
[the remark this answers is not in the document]
eschwartz: I'll plus-one this.

I'm having some trouble understanding what exactly is going on in this controller, and I think some naming tweaks might help.
For example, we have four different "account" references, and it's not immediately clear what they all are:
- `model.Account{}`
- an `AccountData` service
- an `AccountManagerAPI` service
- and `account` package

I kind of wish we had a single account model, and a single service struct for working with this model.

cmd/lambda/accounts/update.go
```diff
@@ -2,116 +2,53 @@ package main
...
-		}
-		// Other DB errors return a 500
-		log.Printf("ERROR: Failed to update account %s: %s", *request.ID, err)
-		response.WriteServerErrorWithResponse(w, "Internal Server Error")
+		api.WriteAPIErrorResponse(w, err)
 		return
 	}
-
-	accountJSON, err := json.Marshal(response.AccountResponse(*acct))
+	err = a.Update(d, amSvc)
```
eschwartz: Does `Update()` mutate the account object? If that's the case, it might just be worth a comment in the method docs.

cmd/lambda/accounts/update.go
```diff
@@ -2,116 +2,53 @@ package main
...
-		)
+	var amSvc accountmanageriface.AccountManagerAPI
+	if err := Services.Config.GetService(&amSvc); err != nil {
+		api.WriteAPIErrorResponse(w, err)
 		return
 	}
 
-	// Update the DB record
-	acct, err := Dao.UpdateAccount(accountPartial, fieldsToUpdate)
+	a, err := account.GetAccountByID(accountID, dataSvc, dataSvc)
```
eschwartz: Do we really need to pass in `dataSvc` twice here?

cmd/lambda/accounts/update.go
```diff
@@ -2,116 +2,53 @@ package main
...
-		accountPartial.Metadata = *request.Metadata
-	}
-	if len(fieldsToUpdate) == 0 {
-		response.WriteRequestValidationError(
-			w,
-			fmt.Sprintf("Unable to update account %s: "+
-				"no updatable fields provided",
-				accountID),
-		)
+	var amSvc accountmanageriface.AccountManagerAPI
```
eschwartz: So back were looking at the account data layer, there was some discussion around the pattern of injecting service dependencies at the method level, versus injecting them on the service structs.

The thing about having to pass in dependencies at the method level is that there's a lot of low-level service configuration and wiring that now needs to happen in the controller itself -- which is how we end up with all these "Account" services in the code, even though I'm really only using one of them.

I'd like if we could get to the point where our controller looked a lot thinner, something just like:

```golang
func UpdateAccountByID(w http.ResponseWriter, r *http.Request) {
	accountID := mux.Vars(r)["accountId"]

	// Deserialize the request JSON as an request object
	accountUpdates := model.Account{}
	decoder := json.NewDecoder(r.Body)
	err := decoder.Decode(&accountUpdates)
	if err != nil {
		api.WriteAPIErrorResponse(w,
			errors.NewBadRequest("invalid request parameters"))
		return
	}

        // Update the account
	updatedAccount, err := config.Accounts.Update(accountUpdates)
	if err != nil {
		api.WriteAPIErrorResponse(w, err)
		return
	}

	api.WriteAPIResponse(w, http.StatusOK, updatedAccount)
}

File: pkg/accountmanager/manager.go (hunk 7 above)
Comment (kddejong) [reply pkg/accountmanager/manager.go]: Yea, this one has been a little tricky.   I think there are some things are the configuration stuff we need to sort out.

File: pkg/accountmanager/manager.go [hunk 12]
```diff
@@ -0,0 +1,74 @@
...
+
+	// Create the credentials from AssumeRoleProvider to assume the role
+	creds := stscreds.NewCredentialsWithClient(am.sts, am.adminRoleArn.String())
+	_, err = creds.Get()
+	if err != nil {
+		return errors.NewValidation("admin role arn", err)
+	}
+	am.awsConfig = aws.NewConfig().WithCredentials(creds)
+
+	am.principalRoleArn, err = arn.Parse(
```
eschwartz: We're making some assumptions here that may not be safe to make.
We actually store the principal role ARN in the Account DynDB table. I could envision some weird bugs popping up here, if the `PRINCIPAL_ROLE_NAME` env var doesn't match whats in the account's DB record

cmd/lambda/accounts/update.go
kddejong: We don't.  The question falls back do we add the smaller interfaces into the struct or do we pass them to the functions that use them.  I prefer the later but there was some concerns on the last PR that we were passing the same stuff a lot.

I could reduce this to one but it could create some weird scenarios for anyone that wanted to use this function.

I would prefer to go back to the later and pass in the interface to the functions that need them.

cmd/lambda/accounts/update.go
kddejong: It changes the data and saves it.  I've been debating if that should be two unique steps.

cmd/lambda/accounts/update_test.go
kddejong: I think you'll need to be more specific.  Some of that logic may have moved someplace else.

pkg/accountmanager/manager.go
eschwartz: > Setup creates a new session manager struct

Is this comment up to date with the code? (what's a "session manager struct", and where is it created?)

cmd/lambda/accounts/update.go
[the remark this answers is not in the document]
kddejong: The are separated for some good reasons.  The naming could be improved I would agree with that.  

I struggle with what that naming could be.  @marinatedpork is looking at some work I've done to remove the model pkg.  Additionally the database save write methods could be made more generic.  The problem we have with making things more generic is there are incosistencies in the data structures from table to table.   With the other changes there may be way to simplify that more.

tests/acceptance/api_test.go
eschwartz: we should be careful about breaking API changes....
Can we at least make sure this is called out in the CHANGELOG, that we're breaking the API interface in the next version?

pkg/accountmanager/manager.go
kddejong: Yea, funny because we probably have that problem all over the place now too.  I think we need to decide how this gets set.  

The environment variable name should be the Default value only?

cmd/lambda/accounts/update.go
kddejong: Yep been trying to stick with the making interfaces smaller mantra.  I actually don't mind the idea of passing in interfaces to functions.   However, since we have greatly simplified the data layer what we have can be considered smaller.  So we can adjust if you would prefer.

cmd/lambda/accounts/update_test.go
eschwartz: talked offline about this -- I was asking specifically about test coverage around which fields are update-able and which aren't. @kddejong says this is covered under the validation logic in `account.Update()`

pkg/accountmanager/manager.go
eschwartz: > The environment variable name should be the Default value only?

Yeah, I'd guess I'd say the env var is used in the create account controller, to decide how to name the principal role/policy in the first place. But then it's persisted to the DB, and should probably be read from the DB in the future.

Or maybe we decide that persisting the role names to the DB is not useful, and drop it. But dong both is not great.

We can punt on this for now, if pulling these from the DB here adds complexity.

PR #215, opened 2020-01-10
Lease data layer

AmudaPalani:
## Proposed changes

Create data layer for leases

## Types of changes

## Checklist


pkg/data/helpers.go
```diff
@@ -37,3 +45,52 @@ func getFiltersFromStruct(i interface{}, keyName *string) (*expression.KeyCondit
...
+
+func Invoke(any interface{}, name string, args... interface{}) error {
+	inputs := make([]reflect.Value, len(args))
+	for i, _ := range args {
+		inputs[i] = reflect.ValueOf(args[i])
+	}
+	result := reflect.ValueOf(any).MethodByName(name).Call(inputs)
+	err := result[0].Interface()
+
+	if err != nil {
```
kddejong: Can you explain this part a little to me?  Seems weird to say err!= nil then return nil for error.

Additional return gErrors.New("error").  Does that mask the original AWS error?  Should we making this an errors.NewInternalServer().

pkg/data/account.go
```diff
@@ -46,38 +43,8 @@ func (a *Account) WriteAccount(account *model.Account, prevLastModifiedOn *int64
...
-	var awsErr awserr.Error
-	if gErrors.As(err, &awsErr) {
-		if awsErr.Code() == "ConditionalCheckFailedException" {
-			return errors.NewConflict(
-				"account",
-				*account.ID,
-				fmt.Errorf("unable to update account: accounts has been modified since request was made"))
-		}
-	}
-	if err != nil {
```
kddejong: This part looks like it was missed in the new function.  We still want the other errors captured and returned.

pkg/data/helpers.go
AmudaPalani: This is WIP. Please ignore this for now.

pkg/data/helpers.go
```diff
@@ -37,3 +45,52 @@ func getFiltersFromStruct(i interface{}, keyName *string) (*expression.KeyCondit
...
+				tableName,
+				"input",
+				fmt.Errorf("unable to update %s: Table has been modified since request was made", tableName))
+		}
+	}
+
+	return nil
+}
+
+func Invoke(any interface{}, name string, args... interface{}) error {
```
joshmarsh: I'm having trouble understanding the purpose of this function. Why not just use normal syntax for calling the PutItem method?
```
	_, err = a.DynamoDB.PutItem(	
		&dynamodb.PutItemInput{	
			// Query in Lease Table	
			TableName: aws.String(a.TableName),	
			// Find Account for the requested accountId	
			Item: putMap.M,	
			// Condition Expression	
			ConditionExpression:       expr.Condition(),	
			ExpressionAttributeNames:  expr.Names(),	
			ExpressionAttributeValues: expr.Values(),	
			// Return the updated record	
			ReturnValues: aws.String(returnValue),	
		},	
	)
```

It looks like the only thing gained here is handling the error inside of this helper function rather than in the calling function

pkg/model/lease.go
```diff
@@ -0,0 +1,89 @@
...
+
+import (
+	"fmt"
+	"strings"
+)
+
+// Lease is a type corresponding to a Lease
+// table record
+type Lease struct {
+	AccountID                string                 `json:"AccountId"`                // AWS Account ID
```
kddejong: Missing the dynamodbav tags.  We are going to need those.

pkg/data/helpers.go
```diff
@@ -37,3 +45,81 @@ func getFiltersFromStruct(i interface{}, keyName *string) (*expression.KeyCondit
...
+	var dataModel *model.Account
+	var i *Account
+
+	if tableName == "account" {
+		i, _ = input.(*Account)
+		dataModel, _ = data.(*model.Account)
+
+	}
+	putMap, _ = dynamodbattribute.Marshal(dataModel)
+	_, err := invoke(i.DynamoDB, "PutItem",
```
kddejong: still not sure I get the value of invoke here.

pkg/data/helpers.go
```diff
@@ -37,3 +45,81 @@ func getFiltersFromStruct(i interface{}, keyName *string) (*expression.KeyCondit
 	}
 	return kb, cb
 }
+
+func putItem(input interface{}, data interface{}, tableName string, expr *expression.Expression) error {
```
kddejong: Another option here is to 
`func putItem(input *dynamodb.PutItemInput) error`

Then WriteAccount can just be building that input and passing it along.

AmudaPalani: Yes, But you have data layer struct type Account struct {} that needs to be made generic to be called DataLayer struct {} and then we can make it completely generic by 

`func putItem(input *dynamodb.PutItemInput, dataInterface *DataLayer) error

pkg/data/account.go
AmudaPalani: Added

pkg/data/helpers.go
AmudaPalani: Hi Josh, I was trying to do error handling inside the putItem. I have made it simpler now.

pkg/model/lease.go
AmudaPalani: Added

pkg/data/helpers.go
AmudaPalani: I was trying to make this method generic to handle errors. But made it simpler now,

pkg/model/lease.go
```diff
@@ -0,0 +1,89 @@
...
+	// LeaseDestroyed means the lease has been deleted via an API call or other user action.
+	LeaseDestroyed LeaseStatusReason = "Destroyed"
+	// LeaseActive means the lease is still active.
+	LeaseActive LeaseStatusReason = "Active"
+	// LeaseRolledBack means something happened in the system that caused the lease to be inactive
+	// based on an error happening and rollback occuring
+	LeaseRolledBack LeaseStatusReason = "Rollback"
+	// AccountOrphaned means that the health of the account was compromised.  The account has been orphaned
+	// which means the leases are also made Inactive
+	AccountOrphaned LeaseStatusReason = "AccountOrphaned"
```
kddejong: Should this be LeaseAccountOrphaned?

pkg/model/lease.go
```diff
@@ -0,0 +1,89 @@
...
+
+import (
+	"fmt"
+	"strings"
+)
+
+// Lease is a type corresponding to a Lease
+// table record
+type Lease struct {
+	AccountID                *string                 `json:"AccountId,omitempty" dynamodbav:"AccountId"`                                         // AWS Account ID
```
kddejong: I think we want the JSON stuff to mimic. 
https://github.com/Optum/dce/blob/master/pkg/api/response/lease.go#L26-L40

Dynamodbav stuff looks great

pkg/data/leases.go
```diff
@@ -0,0 +1,94 @@
...
+	if err != nil {
+		return nil, errors.NewInternalServer("failed unmarshaling of leases", err)
+	}
+	return leases, err
+}
+
+// GetLeases Get a list of leases
+func (a *Account) GetLeases(q *model.Lease) (*model.Leases, error) {
+
+	if q.LeaseStatus != nil {
```
kddejong: The oddness of the lease table is going to come out here.  I think we either need a GetLeaseByID function as well or we can use GetLeases with additional logic to use the Lease ID Index.

pkg/data/lease.go
```diff
@@ -0,0 +1,139 @@
...
+			fmt.Sprintf("delete lease failed for account %q and principal %q", *lease.AccountID, *lease.PrincipalID),
+			err,
+		)
+	}
+
+	return nil
+}
+
+// GetLeaseByID the Lease record by ID
+func (a *Account) GetLeaseByAccountIDAndPrincipalID(accountID string, principalID string) (*model.Lease, error) {
```
kddejong: I wonder if this should just be `GetLeaseByAccountIDAndPrincipalID` I get that this isn't the key index but for the application code maybe we should be treating the Lease ID as the PK.

pkg/data/helpers.go
```diff
@@ -37,3 +38,8 @@ func getFiltersFromStruct(i interface{}, keyName *string) (*expression.KeyCondit
 	}
 	return kb, cb
 }
+
+func putItem(input *dynamodb.PutItemInput, dataInterface *Account) error {
```
kddejong: Still not sure I get the value of this.  Should this exist for query, scan, get, etc.?

pkg/model/lease.go
AmudaPalani: fixed

pkg/data/leases.go
AmudaPalani: Added GetLeaseByID

pkg/data/lease.go
AmudaPalani: Added GetLeaseByID

pkg/data/helpers.go
AmudaPalani: Added query

AmudaPalani: Added get, scan as well

pkg/model/lease.go
AmudaPalani: fixed

PR #217, opened 2020-01-13
Switch accounts list/query to new data model

kddejong:
## Proposed changes

- Update accounts list controllers to user new data model

## Types of changes

## Checklist

### DIFF
--- a/cmd/lambda/accounts/main.go
+++ b/cmd/lambda/accounts/main.go
@@ -80,13 +80,6 @@ var (
 	allowedRegions              []string
 )
 
-const (
-	AccountIDParam     = "id"
-	NextAccountIDParam = "nextId"
-	StatusParam        = "status"
-	LimitParam         = "limit"
-)
-
 func init() {
 	initConfig()
 
@@ -95,21 +88,11 @@ func init() {
 		// Routes with query strings always go first,
 		// because the matcher will stop on the first match
 		api.Route{
-			"GetAccountByStatus",
+			"GetAccounts",
 			"GET",
 			"/accounts",
-			[]string{"accountStatus"},
-			GetAccountByStatus,
-		},
-
-		// Routes without query strings go after all of the
-		// routes that use query strings for matchers.
-		api.Route{
-			"GetAllAccounts",
-			"GET",
-			"/accounts",
-			api.EmptyQueryString,
-			GetAllAccounts,
+			api.EmptyQueryString, // should this be the valid query strings? will this give a 404 then?
+			GetAccounts,
 		},
 		api.Route{
 			"GetAccountByID",
@@ -203,7 +186,7 @@ func Handler(ctx context.Context, req events.APIGatewayProxyRequest) (events.API
 	baseRequest = url.URL{}
 	baseRequest.Scheme = req.Headers["X-Forwarded-Proto"]
 	baseRequest.Host = req.Headers["Host"]
-	baseRequest.Path = req.RequestContext.Stage
+	baseRequest.Path = fmt.Sprintf("%s%s", req.RequestContext.Stage, req.Path)
 
 	// If no name is provided in the HTTP request body, throw an error
 	return muxLambda.ProxyWithContext(ctx, req)
--- a/pkg/data/accounts.go
+++ b/pkg/data/accounts.go
@@ -9,8 +9,13 @@ import (
 	"github.com/aws/aws-sdk-go/service/dynamodb/expression"
 )
 
+type queryScanOutput struct {
+	items            []map[string]*dynamodb.AttributeValue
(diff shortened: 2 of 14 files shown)


pkg/data/accounts.go
```diff
@@ -65,30 +81,64 @@ func (a *Account) scanAccounts(q *model.Account) (*model.Accounts, error) {
...
+		return nil, err
 	}
-	return a.scanAccounts(q)
+
+	q.NextID = nil
+	for _, v := range outputs.lastEvaluatedKey {
+		q.NextID = v.S
+	}
+
+	accounts := &model.Accounts{}
```
kddejong: Moved this code here since its common between both query and scan.

pkg/data/helpers.go
```diff
@@ -13,25 +13,27 @@ func getFiltersFromStruct(i interface{}, keyName *string) (*expression.KeyCondit
...
-		if !reflect.ValueOf(value).IsNil() {
-			switch v.Field(i).Kind() {
-			case reflect.Ptr:
-				dValue := reflect.Indirect(v.Field(i)).Interface()
-				if keyName != nil {
-					if dField == *keyName {
-						newFilter := expression.Key(dField).Equal(expression.Value(dValue))
-						kb = &newFilter
-						continue
+		if dField != "-" {
```
kddejong: Forgot to account on the void tag.  Need this to skip adding queries for limit and nextId

tests/acceptance/api_test.go
```diff
@@ -1870,7 +1870,7 @@ func TestApi(t *testing.T) {
 		t.Run("When there is a status parameter", func(t *testing.T) {
 			resp := apiRequest(t, &apiRequestInput{
 				method: "GET",
-				url:    apiURL + "/accounts?status=" + string(db.NotReady),
+				url:    apiURL + "/accounts?accountStatus=" + string(db.NotReady),
```
kddejong: Match the json payload verbage for account status

pkg/model/account.go
```diff
@@ -2,14 +2,16 @@ package model
...
-	LastModifiedOn      *int64                 `json:"lastModifiedOn,omitempty" dynamodbav:"LastModifiedOn"`                     // Last Modified Epoch Timestamp
-	CreatedOn           *int64                 `json:"createdOn,omitempty"  dynamodbav:"CreatedOn,omitempty"`                    // Account CreatedOn
-	AdminRoleArn        *string                `json:"adminRoleArn,omitempty"  dynamodbav:"AdminRoleArn"`                        // Assumed by the master account, to manage this user account
-	PrincipalRoleArn    *string                `json:"principalRoleArn,omitempty"  dynamodbav:"PrincipalRoleArn,omitempty"`      // Assumed by principal users
-	PrincipalPolicyHash *string                `json:"principalPolicyHash,omitempty" dynamodbav:"PrincipalPolicyHash,omitempty"` // The the hash of the policy version deployed
-	Metadata            map[string]interface{} `json:"metadata,omitempty"  dynamodbav:"Metadata,omitempty"`                      // Any org specific metadata pertaining to the account
+	ID                  *string                `json:"id,omitempty" dynamodbav:"Id" schema:"id,omitempty"`                                                              // AWS Account ID
```
kddejong: schema is needed here.  Different marshaling needs than json.

tests/acceptance/api_test.go
nathanagood: This is a contract (breaking) change. Definitely needs to be called out in CHANGELOG and really should be avoided altogether now that this is public. Almost better to rename the variable....

cmd/lambda/accounts/main.go
```diff
@@ -95,21 +88,11 @@ func init() {
...
-		// Routes without query strings go after all of the
-		// routes that use query strings for matchers.
-		api.Route{
-			"GetAllAccounts",
+			"GetAccounts",
 			"GET",
 			"/accounts",
-			api.EmptyQueryString,
-			GetAllAccounts,
+			api.EmptyQueryString, // should this be the valid query strings? will this give a 404 then?
```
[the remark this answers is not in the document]
kddejong: so as with my comment should I add the values that are queryable here?

tests/acceptance/api_test.go
kddejong: This would be possible to keep it the same.  Its a bummer because the JSON response is accountStatus, the dynamodb is AccountStatus and the query is status.   Also in all fairness the query of account status hasn't been released yet.

https://github.com/Optum/dce/blob/master/pkg/api/response/account.go#L22
https://github.com/Optum/dce/blob/master/pkg/db/model.go#L11

I don't mind calling out a breaking change but it would have been nice to make this accurate instead of perpetuating some of the data silliness we have now.

kddejong:
replaced with #236

PR #230, opened 2020-01-21
allow for deleting a lease by ID

### LINKED ISSUE #121: Delete lease by lease id
**Is your feature request related to a problem? Please describe.**
No

**Describe the solution you'd like**
There is currently no DELETE method implementation on the [/leases/{id} endpoint](https://github.com/Optum/dce/blob/master/modules/swagger.yaml#L470). Users must instead provide principal id and account id to the [/leases endpoint](https://github.com/Optum/dce/blob/master/modules/swagger.yaml#L360) in order to end a lease. Deleting leases by id would be more intuitive.

**Describe alternatives you've considered**

**Additional context**

kddejong:
## Proposed changes
fix #121 

- add the ability to delete leases by ID

## Types of changes

## Checklist

### DIFF
--- a/cmd/lambda/leases/delete.go
+++ b/cmd/lambda/leases/delete.go
@@ -7,6 +7,7 @@ import (
 	"net/http"
 
 	"github.com/Optum/dce/pkg/api/response"
+	"github.com/gorilla/mux"
 
 	"github.com/Optum/dce/pkg/db"
 )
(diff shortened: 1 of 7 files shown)


cmd/lambda/leases/delete_test.go
```diff
@@ -108,6 +113,102 @@ func TestDeleteController_Call(t *testing.T) {
...
+				AccountID:   "123456789012",
+			},
+			transitiionedLease: &db.Lease{
+				ID:          "abc123",
+				LeaseStatus: db.Inactive,
+				PrincipalID: "principal",
+				AccountID:   "123456789012",
+			},
+			getErr: fmt.Errorf("failure"),
+		},
```
AmudaPalani: Can you please add a test for more than one lease exists with input lease ID and in this case DB error should be expected.

cmd/lambda/leases/delete.go
```diff
@@ -36,58 +37,85 @@ func DeleteLease(w http.ResponseWriter, r *http.Request) {
 	log.Printf("Destroying lease %s for Principal %s", accountID, principalID)
 
 	// Move the account to decommissioned
-	accts, err := dao.FindLeasesByPrincipal(principalID)
+	lease, err := dao.GetLease(accountID, principalID)
```
eschwartz: > // Move the account to decommissioned

I think that comment needs to change

cmd/lambda/leases/delete_test.go
kddejong: I'll add a test but it wouldn't matter here.  This logic is done in the db package.  Which will return an error if the result is less than 1 or if the result is more than 1.

I have added a test to the db package.

PR #234, opened 2020-01-21
Don't 404 on accounts GET

kddejong:
## Proposed changes

- No more 404s on lease GETs

## Types of changes

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist

### DIFF
--- a/cmd/lambda/leases/get.go
+++ b/cmd/lambda/leases/get.go
@@ -49,14 +49,14 @@ func GetLeasesByPrincipcalIDAndAccountID(w http.ResponseWriter, r *http.Request)
 		response.WriteServerErrorWithResponse(w, errMsg)
 		return
 	}
-	if lease == nil {
-		log.Printf("Error Getting Lease for Id: %s", err)
-		response.WriteNotFoundError(w)
-		return
+	leaseResponses := []*response.LeaseResponse{}
+	if lease != nil {
+		// Serialize them for the JSON response.
+		leaseResponse := response.LeaseResponse(*lease)
+		leaseResponses = append(leaseResponses, &leaseResponse)
 	}
 
-	leaseResponse := response.LeaseResponse(*lease)
-	err = json.NewEncoder(w).Encode(leaseResponse)
+	err = json.NewEncoder(w).Encode(leaseResponses)
 	if err != nil {
 		errMsg := fmt.Sprintf("Error getting lease for principal %s and acccount %s: %s", principalID, accountID, err.Error())
 		log.Println(errMsg)
--- a/cmd/lambda/leases/get.go
+++ b/cmd/lambda/leases/get.go
@@ -77,11 +77,6 @@ func GetLeasesByPrincipalID(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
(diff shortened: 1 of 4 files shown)


tests/acceptance/api_test.go
```diff
@@ -1559,6 +1559,17 @@ func TestApi(t *testing.T) {
...
 
+		t.Run("When there is a principal ID and an Account ID parameter", func(t *testing.T) {
+			resp := apiRequest(t, &apiRequestInput{
+				method: "GET",
+				url:    apiURL + "/leases?principalId=" + principalIDOne + "&accountId=" + accountIDOne,
+				json:   nil,
+			})
+
+			results := parseResponseArrayJSON(t, resp)
+			assert.Equal(t, 1, len(results), "only one lease should be returned")
```
eschwartz: do we have test coverage for empty results? Might be a good idea...

cmd/lambda/leases/get.go
```diff
@@ -49,14 +49,14 @@ func GetLeasesByPrincipcalIDAndAccountID(w http.ResponseWriter, r *http.Request)
 		response.WriteServerErrorWithResponse(w, errMsg)
 		return
 	}
-	if lease == nil {
-		log.Printf("Error Getting Lease for Id: %s", err)
-		response.WriteNotFoundError(w)
-		return
+	leaseResponses := []*response.LeaseResponse{}
```
nathanagood: So one object was coming back but now an array is coming back. Does that break the contract?

tests/acceptance/api_test.go
nathanagood: It was a single object but is now an array of one. Is this really a valid test that asserts same-as-before-the-break functionality?

cmd/lambda/leases/get.go
eschwartz: Yes. Though I should point out that at one point we _were_ returning empty lists instead of a 404, (as of v0.23.0, at least). So we more recently broke the contract, this change is putting it back.

eschwartz: sorry, I got confused here. The API that broke at some point was the `GET /leases?principalId=`
we'd always had `GET /leases?principalId=&accountId=` return a single object (or 404),
as legacy behavior from before we had Lease IDs.

But I would like to change that, and I'm ok calling that a breaking change to our API.

tests/acceptance/api_test.go
AmudaPalani: Can we add a test to verify an empty array is returned when no record found?

tests/acceptance/api_test.go
```diff
@@ -1559,6 +1559,28 @@ func TestApi(t *testing.T) {
...
+		})
+
+		t.Run("When there is no leases found with principal and account don't exist", func(t *testing.T) {
+			resp := apiRequest(t, &apiRequestInput{
+				method: "GET",
+				url:    apiURL + "/leases?principalId=reallybadprincipal&accountId=notanaccount",
+				json:   nil,
+			})
+
+			results := parseResponseArrayJSON(t, resp)
```
eschwartz: 👍

PR #236, opened 2020-01-22
Convert account list API to use new data model

kddejong:
## Proposed changes

- Update `GET /accounts` to allow for querying with `adminRoleArn`, `principalRoleArn`, and `principalPolicyHash`
- Update `GET /accounts` to use the new data model
- **BREAKING CHANGE:** Change `GET /accounts` query parameter from `status` to `accountStatus` making it consistent with the payload

## Types of changes

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist


pkg/account/service.go
```diff
@@ -153,13 +153,6 @@ func (a *Service) Delete(data *Account) error {
 
 // List Get a list of accounts based on Principal ID
 func (a *Service) List(query *Account) (*Accounts, error) {
-	err := validation.ValidateStruct(query,
```
kddejong: Going to let people list by ID even though it should probably only be a get

pkg/data/mocks/AccountData.go
```diff
@@ -1,54 +0,0 @@
-// Code generated by mockery v1.0.0. DO NOT EDIT.
```
kddejong: Shouldn't have had the mocks here.  My mistake

pkg/data/helpers.go
```diff
@@ -15,25 +15,26 @@ func getFiltersFromStruct(i interface{}, keyName *string) (*expression.KeyCondit
...
-		if !reflect.ValueOf(value).IsNil() {
-			switch v.Field(i).Kind() {
-			case reflect.Ptr:
-				dValue := reflect.Indirect(v.Field(i)).Interface()
-				if keyName != nil {
-					if dField == *keyName {
-						newFilter := expression.Key(dField).Equal(expression.Value(dValue))
-						kb = &newFilter
-						continue
+		if dField != "-" {
```
kddejong: Forgot to filter out Dynamodbav tags that were skips.

pkg/data/accounts.go
```diff
@@ -65,30 +79,66 @@ func (a *Account) scan(query *account.Account) (*account.Accounts, error) {
...
+	}, nil
 }
 
 // List Get a list of accounts
 func (a *Account) List(query *account.Account) (*account.Accounts, error) {
 
+	var outputs *queryScanOutput
+	var err error
+
+	if query.Limit == nil {
```
kddejong: Set the default query limit if one isn't set.

pkg/config/config.go
```diff
@@ -258,13 +258,13 @@ func getKeyPtrs(aMap map[string]ParameterStoreVal) []*string {
 	return keys
 }
 
-// AccountSvc returns the account Service for you
-func (config *ConfigurationBuilder) AccountSvc() accountiface.Servicer {
+// AccountService returns the account Service for you
+func (config *ConfigurationBuilder) AccountService() accountiface.Servicer {
```
kddejong: Spelling out service as this seems more consistent with other things.

cmd/lambda/accounts/list_test.go
```diff
@@ -1,173 +1,125 @@
...
-
-		nextURL := response.BuildNextURL(request, nextParams, url.URL{})
-
-		assert.Equal(t, url.Values{
-			"limit": {
-				"2",
+			retAccounts: &account.Accounts{},
+			retErr:      nil,
+		},
+		{
```
AmudaPalani: Can you please add tests to get accounts with other query parameters accountStatus, adminRoleArn etc,

cmd/lambda/accounts/update_test.go
```diff
@@ -20,6 +20,11 @@ func ptrString(s string) *string {
 	return &ptrS
 }
 
+func ptr64(i int64) *int64 {
+	ptrI := i
+	return &ptrI
+}
+
 func TestUpdateAccountByID(t *testing.T) {
```
AmudaPalani: Can you please add a test for 404- account not found

tests/acceptance/api_test.go
```diff
@@ -1884,7 +1884,7 @@ func TestApi(t *testing.T) {
 		t.Run("When there is a status parameter", func(t *testing.T) {
 			resp := apiRequest(t, &apiRequestInput{
 				method: "GET",
-				url:    apiURL + "/accounts?status=" + string(db.NotReady),
+				url:    apiURL + "/accounts?accountStatus=" + string(db.NotReady),
```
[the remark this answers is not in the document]
AmudaPalani: why do we need to repeat accounts and accountStatus in URL? Can it be simple /accounts/status like it was before?

pkg/account/model.go
```diff
@@ -7,14 +7,16 @@ import (
...
-	AdminRoleArn        *string                `json:"adminRoleArn,omitempty"  dynamodbav:"AdminRoleArn"`                        // Assumed by the master account, to manage this user account
-	PrincipalRoleArn    *string                `json:"principalRoleArn,omitempty"  dynamodbav:"PrincipalRoleArn,omitempty"`      // Assumed by principal users
-	PrincipalPolicyHash *string                `json:"principalPolicyHash,omitempty" dynamodbav:"PrincipalPolicyHash,omitempty"` // The the hash of the policy version deployed
-	Metadata            map[string]interface{} `json:"metadata,omitempty"  dynamodbav:"Metadata,omitempty"`                      // Any org specific metadata pertaining to the account
+	ID                  *string                `json:"id,omitempty" dynamodbav:"Id" schema:"id,omitempty"`                                                              // AWS Account ID
+	Status              *Status                `json:"accountStatus,omitempty" dynamodbav:"AccountStatus,omitempty" schema:"accountStatus,omitempty"`                   // Status of the AWS Account
```
AmudaPalani: Can we change payload to be just status?

tests/acceptance/api_test.go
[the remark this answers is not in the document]
kddejong: for consistency.  I don't mind changing it all to `status` but its weird to me to have `accountStatus` in the json payload and then have to know to use `status` if you are doing a query.  I would like them all to be consistent one way or the other.

pkg/account/model.go
kddejong: We can.  Its a different breaking change.  I would prefer it all be `status`

cmd/lambda/accounts/update_test.go
kddejong: This test doesn't go here and is covered here.  The account Service will either return the updated result or an error.  The error has the details on the coding.  

https://github.com/Optum/dce/blob/6346f4766c7cb05e32dfc4156638500875b171b8/pkg/data/account_test.go#L60-L69

cmd/lambda/accounts/list_test.go
kddejong: These tests are written to handle code paths.  So my question is what are you looking to get from the tests?  Are we validating the Account struct can decode all the different query parameters?

PR #240, opened 2020-01-22
Lease controller - GET

AmudaPalani:
## Proposed changes

Use new data model for leases GET controller

## Types of changes

## Checklist


cmd/lambda/leases/get_test.go
```diff
@@ -1,66 +1,163 @@
...
+	}
+	tests := []struct {
+		name     string
+		expResp  response
+		leaseID  string
+		retLease *lease.Lease
+		retErr   error
+	}{
+		{
+			name:    "success",
```
joshmarsh: Making these more descriptive would make the test easier to parse. It's probably also good to be somewhat consistent with how descriptive the names of the non-table driven tests are below.

cmd/lambda/leases/main.go
```diff
@@ -31,21 +45,33 @@ const (
...
-	maxLeaseBudgetAmount     *float64
-	maxLeasePeriod           *int64
-	defaultLeaseLengthInDays *int
+	// Soon to be deprecated - Legacy support
+	Config             common.DefaultEnvConfig
+	awsSession         *session.Session
+	dao                db.DBer
+	snsSvc             common.Notificationer
+	usageSvc           usage.Service
+	leaseAddedTopicARN string
```
joshmarsh: Aren't some of these also in leaseControllerConfiguration?

cmd/lambda/leases/main.go
```diff
@@ -127,6 +147,53 @@ func init() {
...
+		WithLeaseService().
+		Build()
+	if err != nil {
+		panic(err)
+	}
+
+	Services = svcBldr
+
+	//resetQueueURL = Config.GetEnvVar("RESET_SQS_URL", "DefaultResetSQSUrl")
+	leaseAddedTopicARN = Config.GetEnvVar("LEASE_ADDED_TOPIC", "DCEDefaultProvisionTopic")
```
joshmarsh: Didn't we already retrieve this [when we unmarshalled the env vars into the Settings struct](https://github.com/Optum/dce/pull/240/files#diff-6433aa3e05979c963e2ea37c8e383bc0R154) above?

cmd/lambda/leases/main_test.go
```diff
@@ -0,0 +1,46 @@
...
+
+import (
+	"os"
+	"testing"
+
+	"github.com/aws/aws-lambda-go/events"
+)
+
+func TestMain(m *testing.M) {
+	os.Setenv("LEASE_ADDED_TOPIC", "mock-lease-added-topic")
```
kddejong: Do we need to do this?  Can we just use the defaults?

pkg/config/servicebuild.go
```diff
@@ -291,3 +312,64 @@ func (bldr *ServiceBuilder) createAccountService(config ConfigurationServiceBuil
...
+		return err
+	}
+
+	dataSvcImpl.DynamoDB = dynamodbSvc
+
+	config.WithService(dataSvcImpl)
+	return nil
+}
+
+func (bldr *ServiceBuilder) createLeaseManager(config ConfigurationServiceBuilder) error {
```
kddejong: Thought we were removing this?

pkg/lease/service.go
```diff
@@ -52,12 +52,14 @@ type Eventer interface {
 
 // Manager manages all the actions against a lease
 type Manager interface {
```
kddejong: Lease isn't going to have a manager service so this could be deleted.

pkg/leasemanager/manager.go
```diff
@@ -0,0 +1,24 @@
+package leasemanager
```
kddejong: ditto.  I don't think this needs to exist.

cmd/lambda/leases/main.go
AmudaPalani: For now we are supporting both old and new ways of interacting with db layer.

cmd/lambda/leases/main.go
AmudaPalani: For now we are supporting both old and new ways of interacting with db layer.

cmd/lambda/leases/get_test.go
AmudaPalani: Same as accounts.

pkg/leasemanager/manager.go
AmudaPalani: fixed

pkg/config/servicebuild.go
AmudaPalani: fixed

cmd/lambda/leases/main_test.go
AmudaPalani: fixed

cmd/lambda/leases/get_test.go
AmudaPalani: Modified test names.

pkg/lease/service.go
AmudaPalani: fixed

cmd/lambda/leases/get_test.go
```diff
@@ -1,66 +1,163 @@
...
+		}
+
+		mockRequest := events.APIGatewayProxyRequest{HTTPMethod: http.MethodGet, Path: "/leases/unique-id"}
 
 		actualResponse, err := Handler(context.TODO(), mockRequest)
 		require.Nil(t, err)
 
 		require.Equal(t, actualResponse.StatusCode, 500, "Returns a 500.")
-		require.Equal(t, actualResponse.Body, "{\"error\":{\"code\":\"ServerError\",\"message\":\"Failed Get on Lease unique-id\"}}", "Returns an error response.")
+		require.Equal(t, MockAPIErrorResponse(http.StatusInternalServerError, "{\"error\":{\"message\":\"unknown error\",\"code\":\"ServerError\"}}\n"), actualResponse, "Returns an error response.")
```
kddejong: I would prefer what is in the accounts controller here.
```
assert.Equal(t, tt.expResp.StatusCode, resp.StatusCode)
assert.Equal(t, tt.expResp.Body, string(body))
```
or even better would probably be using https://godoc.org/github.com/stretchr/testify/assert#JSONEq

cmd/lambda/leases/main.go
```diff
@@ -127,6 +147,52 @@ func init() {
...
+
+	_, err = svcBldr.
+		// AWS services...
+		WithDynamoDB().
+		WithSTS().
+		WithS3().
+		WithSNS().
+		WithSQS().
+		// DCE services...
+		WithStorageService().
```
kddejong: I'm not sure we need this.  I would prefer to add it back as we use it?  But nothing too critical here.

pkg/config/config.go
```diff
@@ -268,3 +269,14 @@ func (config *ConfigurationBuilder) AccountSvc() accountiface.Servicer {
 
 	return accountSvc
 }
+
+// LeaseSvc returns the lease Service for you
+func (config *ConfigurationBuilder) LeaseSvc() leaseiface.Servicer {
```
kddejong: I'm working to rename this to AccountService so we should probably change this to LeaseService.

cmd/lambda/leases/get_test.go
AmudaPalani: fixed

cmd/lambda/leases/main.go
AmudaPalani: fixed

pkg/config/config.go
AmudaPalani: renamed

PR #246, opened 2020-01-27
New Data Model for the Account Controller Create Method

kddejong:
## Proposed changes

- Update create account controller to use the new data model

## Types of changes

## Checklist


cmd/lambda/accounts/create_test.go
```diff
@@ -1,686 +1,95 @@
...
-			}),
-		).Return(&rolemanager.CreateRoleWithPolicyOutput{}, nil)
-
-		// Call the controller with the account
-		_, err := Handler(
-			context.TODO(),
-			createAccountAPIRequest(t, CreateRequest{
-				ID:           "1234567890",
-				AdminRoleArn: "arn:mock",
+			name: "success",
```
AmudaPalani: Can you please update the description to Gherkins style? I updated in lease to "When lease service - Get returns success".

cmd/lambda/accounts/main_test.go
```diff
@@ -1,11 +1,9 @@
 package main
 
 import (
-	"encoding/json"
 	"os"
 	"testing"
 
-	"github.com/Optum/dce/pkg/api/response"
 	"github.com/aws/aws-lambda-go/events"
 )
```
AmudaPalani: Can you please add tests to run through the main's handler and init methods to call the create?

pkg/account/model.go
```diff
@@ -1,22 +1,50 @@
...
+	StatusNone,
+	StatusLeased,
+	StatusNotReady,
+	StatusOrphaned,
+	StatusReady,
+}
+
+func init() {
+	var ok bool
+	PrincipalPolicyName, ok = os.LookupEnv("PRINCIPAL_POLICY_NAME")
```
joshmarsh: Why not inject this?

kddejong: That's not easily injectable.
1. As opposed to the Principal Role Name we don't store the policy name in the database.
2. When unmarshaling JSON/DynamoDB I want to populate this value so we can pass it along with the account.  That means I need it outside the struct and I need it some place those functions can get at it.

pkg/account/service.go
```diff
@@ -133,6 +138,68 @@ func (a *Service) Update(ID string, data *Account) (*Account, error) {
...
+		ID:                *data.ID,
+		AdminRoleArn:      *data.AdminRoleArn,
+		Metadata:          data.Metadata,
+		PrincipalRoleName: a.principalRoleName,
+	})
+	if err != nil {
+		return nil, err
+	}
+
+	err = a.managerSvc.MergePrincipalAccess(new)
```
joshmarsh: Should we update this to UpsertPrincipalAccess?

pkg/account/service.go
```diff
@@ -47,19 +48,24 @@ type ReaderWriterDeleter interface {
...
 
 // Manager manages all the actions against an account
 type Manager interface {
-	Setup(adminRole string) error
+	ValidateAccess(role *arn.ARN) error
+	MergePrincipalAccess(account *Account) error
 }
 
 // Service is a type corresponding to a Account table record
 type Service struct {
```
joshmarsh: It seems like 'AccountService' would be more accurate.

cmd/lambda/accounts/main_test.go
kddejong: Done

pkg/account/service.go
kddejong: Done

PR #248, opened 2020-01-29
Remove extra hopes in configuring principal policy

kddejong:
## Proposed changes

Right now to add the IAM Deny Policy tags and allowed regions to the principal policy we have a go template in the s3 bucket.  Then we configure environment variables in Lambda to set those values at implementation time.

This creates a few issues but the most important being that the etag will not change when the environment variables are changed.  Meaning a policy will not be reapplied if the environment variables are changed and the policy in the bucket stays the same.  Since we are putting those static values directly into the template we skip the extra layer of the lambda to do the work making things simpler and more reliable.   The etag will now change when those variables are changed.

## Types of changes

## Checklist


kddejong:
Going to close this.  We will have to address this issue a different way.

PR #251, opened 2020-01-29
switch from strings to json for the sqs reset queue

kddejong:
## Proposed changes

- Switch to using JSON messages of the account records when putting messages into the SQS Reset Queue

This will get easier with the other rewrites but for now this is the simplest of the changes to separate the string to json portion.

## Types of changes

## Checklist

### DIFF
--- a/cmd/lambda/accounts/create.go
+++ b/cmd/lambda/accounts/create.go
@@ -105,7 +105,12 @@ func CreateAccount(w http.ResponseWriter, r *http.Request) {
 	}
 
 	// Add Account to Reset Queue
-	err = Queue.SendMessage(&resetQueueURL, &account.ID)
+	body, err := json.Marshal(&account)
+	if err != nil {
+		log.Printf("Failed to add account %s to reset Queue: %s", account.ID, err)
+	}
+	sBody := string(body)
+	err = Queue.SendMessage(&resetQueueURL, &sBody)
 	if err != nil {
 		log.Printf("Failed to add account %s to reset Queue: %s", account.ID, err)
 		response.WriteServerErrorWithResponse(w, "Internal server error")
--- a/cmd/lambda/leases/main.go
+++ b/cmd/lambda/leases/main.go
@@ -23,7 +23,6 @@ import (
 
 type leaseControllerConfiguration struct {
 	Debug                    string  `env:"DEBUG" defaultEnv:"false"`
-	ResetQueueURL            string  `env:"RESET_SQS_URL" defaultEnv:"DefaultResetSQSUrl"`
 	LeaseAddedTopicARN       string  `env:"LEASE_ADDED_TOPIC" defaultEnv:"DCEDefaultProvisionTopic"`
 	DecommissionTopicARN     string  `env:"DECOMMISSION_TOPIC" defaultEnv:"DefaultDecommissionTopicArn"`
 	CognitoUserPoolID        string  `env:"COGNITO_USER_POOL_ID" defaultEnv:"DefaultCognitoUserPoolId"`
@@ -63,7 +62,6 @@ var (
 	usageSvc           usage.Service
 	leaseAddedTopicARN string
 	//decommissionTopicARN     string
-	//resetQueueURL            string
 	principalBudgetAmount    float64
 	principalBudgetPeriod    string
 	maxLeaseBudgetAmount     float64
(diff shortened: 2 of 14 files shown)


cmd/lambda/accounts/create.go
```diff
@@ -105,7 +105,13 @@ func CreateAccount(w http.ResponseWriter, r *http.Request) {
 	}
 
 	// Add Account to Reset Queue
-	err = Queue.SendMessage(&resetQueueURL, &account.ID)
+	body, err := json.Marshal(&account)
```
kddejong: This plus the next 4-5 lines will go away as we make more changes.  This is repeated here as it is the best way to do it at this point.

cmd/lambda/leases/main.go
```diff
@@ -23,7 +23,6 @@ import (
 
 type leaseControllerConfiguration struct {
 	Debug                    string  `env:"DEBUG" defaultEnv:"false"`
-	ResetQueueURL            string  `env:"RESET_SQS_URL" defaultEnv:"DefaultResetSQSUrl"`
```
kddejong: Not being used anywhere in this controller.  Just adds confusion.

pkg/processresetqueue/reset.go
```diff
@@ -19,7 +19,6 @@ type ResetInput struct {
 	ResetQueueURL *string
 	ResetBuild    common.Builder
 	BuildName     *string
-	DbSvc         db.DBer
```
kddejong: We no longer need to read from the DB as the entire record is in the queue.

pkg/processresetqueue/reset.go
```diff
@@ -68,51 +67,27 @@ func Reset(input *ResetInput) (*ResetOutput, error) {
...
 			result := ResetResult{
 				BuildTrigger:    false,
 				MessageDeletion: false,
 			}
 
-			// Get the Account from the Database
-			account, err := input.DbSvc.GetAccount(accountID)
-			if err != nil {
-				failTriggerResetOnAccount(&output, result, accountID,
+			acct := &account.Account{}
```
kddejong: Using the new account model as this will parse things into the ARNs and make the next set of lines of code unnecessary .

pkg/processresetqueue/reset_test.go
```diff
@@ -334,157 +262,9 @@ func TestReset(t *testing.T) {
 			require.Equal(t, test.ExpectedError, err)
 
 			// Assert that mockBuild was called as expected
-			mockBuilder.AssertNumberOfCalls(t, "StartBuild", test.ExpectedBuildCount)
-		}
-	})
-
-	t.Run("Should set Lease.Status=ResetLock in DB, if the account has active leases", func(t *testing.T) {
```
kddejong: The following tests add no value.

cmd/lambda/publish_lease_events/main.go
```diff
@@ -113,18 +113,22 @@ func handleRecord(input *handleRecordInput) error {
 		if didBecomeInactive {
 			// Before adding the account to any queues, make sure the account is
 			// updated to NotReady state.
-			_, err = input.dbSvc.TransitionAccountStatus(lease.AccountID, db.Leased, db.NotReady)
+			acct, err := input.dbSvc.TransitionAccountStatus(lease.AccountID, db.Leased, db.NotReady)
 			if err != nil {
-				log.Printf("ERROR: Failed to mark AccountStatus=NotReady for %s, after lease for %s became inactive",
```
eschwartz: ?

pkg/processresetqueue/reset.go
nathanagood: 👍 Not only do we read the entire record, but it's a rather important point that _we're reading the record as a snapshot at the time the record was written_. This is rather a critical point if we're writing events that can be played back or are idempotent.

pkg/processresetqueue/reset.go
```diff
@@ -68,51 +67,28 @@ func Reset(input *ResetInput) (*ResetOutput, error) {
...
 
 			// Set Reset Build Env Vars
 			resetBuildEnvironment := map[string]string{
-				"RESET_ACCOUNT":                     account.ID,
-				"RESET_ACCOUNT_ADMIN_ROLE_NAME":     accountAdminRoleName,
-				"RESET_ACCOUNT_PRINCIPAL_ROLE_NAME": accountPrincipalRoleName,
+				"RESET_ACCOUNT":                     accountID,
+				"RESET_ACCOUNT_ADMIN_ROLE_NAME":     *acct.AdminRoleArn.IAMResourceName(),
+				"RESET_ACCOUNT_PRINCIPAL_ROLE_NAME": *acct.PrincipalRoleArn.IAMResourceName(),
 			}
```
nathanagood: This seems important that it's done always. Just wondering if it should be in a method that's `defer`ed as soon as we can...

cmd/lambda/publish_lease_events/main.go
nathanagood: @eschwartz are you "?" why the logging statement was removed? @kddejong, does seem like that information might be useful for debugging.

kddejong: Partially.  It will already print the error but juust not there.  So in a way this print is duplicative.  Additionally this error handling doesn't return err it continues on... silently continues if you will.

PR #252, opened 2020-01-30
Use the new data modeling for account creation controller

kddejong:
## Proposed changes

- Migrate account create functionality to the new data model

## Types of changes

## Checklist


cmd/lambda/accounts/create.go
```diff
@@ -2,251 +2,31 @@ package main
...
-	if err != nil {
-		log.Printf("Failed to publish SNS account-created message for %s: %s", account.ID, err)
-		response.WriteServerErrorWithResponse(w, "Internal server error")
+		api.WriteAPIErrorResponse(w,
+			errors.NewBadRequest("invalid request parameters"))
 		return
 	}
 
-	accountResponseJSON, err := json.Marshal(accountResponse)
+	account, err := Services.Config.AccountService().Create(newAccount)
```
eschwartz: I like this short syntax for the config builder 👍

cmd/lambda/accounts/create.go
```diff
@@ -2,251 +2,31 @@ package main
...
-	})
-	return createRoleOutput, policyHash, err
-}
-
-type principalPolicyInput struct {
-	PrincipalPolicyArn   string
-	PrincipalRoleArn     string
-	PrincipalIAMDenyTags []string
-	AdminRoleArn         string
-	Regions              []string
```
eschwartz: Does it keep you up at night, @kddejong, the number of lines of code you've killed with your bare hands? 😜

pkg/account/service.go
```diff
@@ -133,6 +138,68 @@ func (a *Service) Update(ID string, data *Account) (*Account, error) {
...
+	if err != nil {
+		return nil, err
+	}
+
+	err = a.Save(new)
+	if err != nil {
+		return nil, err
+	}
+
+	err = a.eventSvc.AccountCreate(new)
```
eschwartz: I like this setup, but I think the difference of intention between `eventSvc.AccountCreate()` and `eventSvc.AccountReset()` is a little fuzzy.

For `AccountCreate`, we're saying that we _did_ create an account, for anyone who cares
For `AccountReset`, we're saying that we _want to_ reset the account.

It's just a naming change, but I think it would be more clear as something `AccountCreated`(passive) / `ResetAccount` (imperative)

Or I've even setup event topics like `DidCreateAccount` / `PleaseResetAccount`, to be extra explicit.

eschwartz: would also be worth an inline command on line 195, like "this adds the account to the reset queue, to trigger nuke to run on the account"

pkg/accountmanager/service.go
```diff
@@ -0,0 +1,135 @@
...
+
+// Service manages account resources
+type Service struct {
+	client   clienter
+	storager common.Storager
+	config   ServiceConfig
+}
+
+// ValidateAccess creates a new Account instance
+func (s *Service) ValidateAccess(role *arn.ARN) error {
```
eschwartz: I like the rename from `Setup` --> `ValidateAccess` 👍

pkg/accountmanager/service.go
```diff
@@ -0,0 +1,135 @@
...
+	PrincipalRoleDescription    string   `env:"PRINCIPAL_ROLE_DESCRIPTION" envDefault:"Role for principal users of DCE"`
+	PrincipalPolicyDescription  string   `env:"PRINCIPAL_POLICY_DESCRIPTION" envDefault:"Policy for principal users of DCE"`
+	tags                        []*iam.Tag
+	assumeRolePolicy            string
+}
+
+// Service manages account resources
+type Service struct {
+	client   clienter
+	storager common.Storager
```
eschwartz: why not through `storager` into the `clienter` structs -- seems like that's where AWS service clients are being held

pkg/accountmanager/client.go
```diff
@@ -0,0 +1,58 @@
...
+
+	// return no config for nil inputs
+	if roleArn == nil {
+		return nil
+	}
+
+	key := roleArn.String()
+
+	// check for cached config
+	if c.configs != nil && c.configs[key] != nil {
```
eschwartz: This caching makes me a little nervous, that we're going to have some invalid cached values somehow.
Given that we're running in Lambda, is this really worth it?

pkg/accountmanager/iam.go
```diff
@@ -0,0 +1,184 @@
+package accountmanager
+
```
eschwartz: this is adapted from existing code, no?

pkg/accountmanager/client.go
kddejong: The value of these credentials is they auto-renew.  We are storing the thing that manages the auto-renewal.  Helps prevent the system from assuming roles every time we take action.

pkg/accountmanager/iam.go
kddejong: Yes, basically reworking what is in the policy and role manager

cmd/lambda/accounts/create.go
eschwartz: Seriously though, this is really cool, how clean this controller becomes

pkg/accountmanager/client.go
eschwartz: That's what I figured. I just always get anxious about caching these kind of things.

pkg/accountmanager/service.go
kddejong: Good question.  I'm using the clienter as a way to keep access to some aws configs that we can share across clients.

pkg/account/service.go
kddejong: Yea let me see if I can find some good naming for this.  Good call.

Issue #165, opened 2019-12-12
Use pkg/config to configure Lambdas 


kddejong:
The new pkg/config exists in master but needs to be implemented for all appropriate resources.  

- [ ] cmd/lambda/accounts
- [ ] cmd/lambda/fan_out_update_lease_status
- [ ] cmd/lambda/lease_auth
- [ ] cmd/lambda/leases
- [ ] cmd/lambda/populate_reset_queue
- [ ] cmd/lambda/process_reset_queue
- [ ] cmd/lambda/publish_lease_events_queue
- [ ] cmd/lambda/update_lease_status
- [ ] cmd/lambda/update_principal_policy
- [ ] cmd/lambda/usage
- [ ] cmd/codebuild/reset

PR #258, opened 2020-01-31
move account delete functionality to new data model

kddejong:
## Proposed changes

- Move Account Deletion into the new data model

## Types of changes

## Checklist


pkg/api/error.go
```diff
@@ -31,10 +36,12 @@ func WriteAPIErrorResponse(w http.ResponseWriter, err error) {
 // WriteAPIResponse writes the response out to the provided ResponseWriter
 func WriteAPIResponse(w http.ResponseWriter, status int, body interface{}) {
 	w.WriteHeader(status)
-	err := json.NewEncoder(w).Encode(body)
-	if err != nil {
-		log.Printf("error encoding and writing message: %+v", body)
-		w.WriteHeader(http.StatusInternalServerError)
-		_ = json.NewEncoder(w).Encode(errors.NewInternalServer("error writing response", err))
+	if body != nil {
```
kddejong: Don't write the body if the interface is nil

pkg/data/account.go
```diff
@@ -87,7 +87,7 @@ func (a *Account) Delete(account *account.Account) error {
 			// Query in Lease Table
 			TableName: aws.String(a.TableName),
 			// Return the updated record
-			ReturnValues: aws.String("ALL_NEW"),
+			ReturnValues: aws.String("NONE"),
```
kddejong: Can't be ALL_NEW.  Wasn't using this before so found the issue now.

pkg/account/model.go
```diff
@@ -90,6 +92,32 @@ func (a *Account) UnmarshalJSON(data []byte) error {
 	return nil
 }
 
+// UnmarshalDynamoDBAttributeValue handles custom unmarshaling of an ARN
+func (a *Account) UnmarshalDynamoDBAttributeValue(av *dynamodb.AttributeValue) error {
+
+	type Alias Account
```
kddejong: If anyone is wondering why this exists like this.  Its to prevent infinite loops with line 101.

kddejong:
Completes the rewrite for #165 for the accounts controller

cmd/lambda/accounts/delete.go
```diff
@@ -1,113 +1,28 @@
...
-	// json.NewEncoder(w).Encode(response.CreateAPIGatewayResponse(http.StatusNoContent, ""))
-	response.WriteAPIResponse(w, http.StatusNoContent, "")
-}
-
-// sendSNS sends notification to SNS that the delete has occurred.
-func sendSNS(account *db.Account) {
-	serializedAccount := response.AccountResponse(*account)
-	serializedMessage, err := common.PrepareSNSMessageJSON(serializedAccount)
 
+	acct, err := Services.Config.AccountService().Get(accountID)
```
eschwartz: Might be worth calling out in the `Get()` method what kind of errors it returns
(I just had to dig in a bit to figure out that we're actually generating different HTTP error types at the data layer)

pkg/account/model.go
eschwartz: Is this `UnmarshalDynamoDBAttributeValue()` method used anywhere? My IDE couldn't find any usages...

pkg/accountmanager/service.go
```diff
@@ -86,6 +86,42 @@ func (s *Service) UpsertPrincipalAccess(account *account.Account) error {
 	return nil
 }
 
+// DeletePrincipalAccess removes all the principal roles and policies
+func (s *Service) DeletePrincipalAccess(account *account.Account) error {
```
eschwartz: Is this new behavior?

pkg/account/service.go
```diff
@@ -215,6 +218,26 @@ func (a *Service) Delete(data *Account) error {
 		return err
 	}
 
+	err = a.managerSvc.DeletePrincipalAccess(data)
+	if err != nil {
+		return err
+	}
+
+	err = a.dataSvc.Delete(data)
```
eschwartz: Looks like this is a duplicate call, see line 216

pkg/account/model.go
kddejong: yes, its an interface used by dynamodbattribute.Unmarshal

pkg/account/service.go
kddejong: Yikes good catch.

pkg/accountmanager/service.go
kddejong: Moved code from role and policy manager.

PR #260, opened 2020-02-02
silently continue when account status change fails

kddejong:
## Proposed changes

- Revert back a change that was made to silently continue when the lease status changes to inactive and the lease event lambda was failing trying to change it.

## Types of changes

## Checklist


joshmarsh:
I see that this is an exact revert of https://github.com/Optum/dce/commit/6fc6d1ec760c5a8a23c27aa2886bfa6ece063114#diff-b470999e9ff9fca0d3d8e19932d813bcL118.

Whats the reason for reverting this back?

kddejong:
We caused an issue because lease deletion was also switching the account status.  We were always failing here because the account status was already set correctly.  I switched it back so the process would silently continue.

PR #262, opened 2020-02-04
Populate reset queue with not ready accounts

kddejong:
## Proposed changes

- Switch account reset population to use the new data model 
- Allow for some pagination on populating the account reset

## Types of changes

## Checklist


eschwartz:
@kddejong taking a 👀 now

cmd/lambda/populate_reset_queue/main.go
```diff
@@ -1,80 +1,85 @@
...
+	cfgBldr := &config.ConfigurationBuilder{}
+	settings = &configuration{}
+	if err := cfgBldr.Unmarshal(settings); err != nil {
+		log.Fatalf("Could not load configuration: %s", err.Error())
 	}
 
-	// Get NotReady Accounts
-	accounts, err := dbSvc.FindAccountsByStatus(db.NotReady)
+	// load up the values into the various settings...
+	err := cfgBldr.WithEnv("AWS_CURRENT_REGION", "AWS_CURRENT_REGION", "us-east-1").Build()
```
eschwartz: not necessarily a change for this PR, but this is still a little awkward to me how we're pulling in env vars like this, 
which are actually hidden requirements by our DB service.

It seems that if our DB service needs this env var, it's own service constructor should be in charge of declaring that.
I want our controller to only have to declare the env vars it's using directly.

cmd/lambda/populate_reset_queue/main.go
```diff
@@ -1,80 +1,85 @@
...
+
+	services = svcBldr
+
+}
+
+// Handler is the base handler function for the lambda
+func Handler(cloudWatchEvent events.CloudWatchEvent) error {
+
+	var eventSvc eventiface.Servicer
+	if err := services.Config.GetService(&eventSvc); err != nil {
```
eschwartz: Didn't we do something in another controller, just like:

```golang
eventSvc := service.GetEventer()
```

that's a nicer pattern, IMO, if we can do it.

kddejong: Yea I think we just need to decide how far down that rabbit hole we want to go.  When do we decide if a service is going to have the easier pattern.

cmd/lambda/populate_reset_queue/main.go
kddejong: I agree with this.  Would be a nice fix.  No repetition if its always needed.

pkg/account/service.go
```diff
@@ -229,6 +230,27 @@ func (a *Service) List(query *Account) (*Accounts, error) {
 	return accounts, nil
 }
 
+// Reset initiates the Reset account process.  It will not change the status as there may
+// be many reasons why a reset is called.  Delete, Lease Ending, etc.
+func (a *Service) Reset(data *Account) error {
```
eschwartz: 👍

PR #270, opened 2020-02-05
usage data layer

kddejong:
## Proposed changes

- Add new data layer for usage

## Types of changes

## Checklist

### DIFF
--- a/cmd/lambda/usage/list.go
+++ b/cmd/lambda/usage/list.go
@@ -35,7 +35,15 @@ func GetUsage(w http.ResponseWriter, r *http.Request) {
 	// Serialize them for the JSON response.
 	usageResponseItems := []response.UsageResponse{}
 	for _, usageItem := range result.Results {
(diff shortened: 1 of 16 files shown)


cmd/lambda/usage/list.go
```diff
@@ -35,7 +35,15 @@ func GetUsage(w http.ResponseWriter, r *http.Request) {
 	// Serialize them for the JSON response.
 	usageResponseItems := []response.UsageResponse{}
 	for _, usageItem := range result.Results {
-		usageResponseItems = append(usageResponseItems, response.UsageResponse(*usageItem))
+		usageResponseItems = append(usageResponseItems, response.UsageResponse{
```
eschwartz: Do we need a `response.UsageResponse` struct? Or can we just use the `usage.Usage` model struct?

kddejong: Yea @eschwartz Going to work on that next.  Little bit different of a change here because pkg/usage already existed so I needed to rework some stuff first.

PR #273, opened 2020-02-06
fix an issue with having the account record

kddejong:
## Proposed changes

- Fix an issue when transitionaccountstatus fails and we don't get the account record.  Make a separate call to get the account so we know we have the correct record.

## Types of changes

## Checklist

### DIFF
--- a/cmd/lambda/publish_lease_events/main.go
+++ b/cmd/lambda/publish_lease_events/main.go
@@ -113,10 +113,19 @@ func handleRecord(input *handleRecordInput) error {
 		if didBecomeInactive {
 			// Before adding the account to any queues, make sure the account is
 			// updated to NotReady state.
+			var acct *db.Account
 			acct, err := input.dbSvc.TransitionAccountStatus(lease.AccountID, db.Leased, db.NotReady)
 			if err != nil {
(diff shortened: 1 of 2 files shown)


cmd/lambda/publish_lease_events/main.go
```diff
@@ -113,12 +113,18 @@ func handleRecord(input *handleRecordInput) error {
...
 			// Before adding the account to any queues, make sure the account is
 			// updated to NotReady state.
-			acct, err := input.dbSvc.TransitionAccountStatus(lease.AccountID, db.Leased, db.NotReady)
+			_, err := input.dbSvc.TransitionAccountStatus(lease.AccountID, db.Leased, db.NotReady)
 			if err != nil {
 				log.Printf("ERROR: Failed to mark AccountStatus=NotReady for %s, after lease for %s became inactive",
 					lease.AccountID, lease.PrincipalID)
 			}
 
+			acct, err := input.dbSvc.GetAccount(lease.AccountID)
```
eschwartz: Not a big deal, but you could probably throw this inside the `if err != nil {}` block, so it only does an additional query if the transition query fails

kddejong: Very true.  I'm going to fix it quick.  Saving DB calls is what I would like to do.

PR #279, opened 2020-02-10
Move service getters to ServiceBuilder struct

eschwartz:
## Proposed changes

Just moving the the configuration service's `LeaseService()` and `AccountService()` getter methods to the main `ServiceBuilder` struct (were in the `ConfigurationBuilder` struct before)
The actual methods for defining these service (`WithLeaseService()`/`WithAccountService()`) are in the `ServiceBuilder`, so it didn't make much sense to have the getter methods in a different struct.

Ran into this while trying to define a new service, and was trying to understand the existing patterns for doing so.

## Types of changes

## Checklist

- [ ] N/A ~I have updated the `CHANGELOG.md` under a `## next` release, with a short summary of my changes~

### DIFF
--- a/cmd/lambda/accounts/create.go
+++ b/cmd/lambda/accounts/create.go
@@ -22,7 +22,7 @@ func CreateAccount(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	account, err := Services.Config.AccountService().Create(newAccount)
+	account, err := Services.AccountService().Create(newAccount)
 	if err != nil {
 		api.WriteAPIErrorResponse(w, err)
 		return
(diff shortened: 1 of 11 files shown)


cmd/lambda/leases/main.go
```diff
@@ -128,13 +128,10 @@ func initConfig() {
 	}
 
 	// load up the values into the various settings...
-	err := cfgBldr.WithEnv("AWS_CURRENT_REGION", "AWS_CURRENT_REGION", "us-east-1").Build()
-	if err != nil {
-		log.Printf("Error: %+v", err)
-	}
+	_ = cfgBldr.WithEnv("AWS_CURRENT_REGION", "AWS_CURRENT_REGION", "us-east-1").Build()
```
eschwartz: this was actually a mistake in the code here -- `WithEnv` returns the `ConfigurationBuilder` object, not an error (so this would have logged an error message with every run)

kddejong:
Thanks for separating this.

PR #283, opened 2020-02-12
Dont' check lease status on upsert principal access

kddejong:
## Proposed changes

- Don't validate status on upsert principal access

## Types of changes

## Checklist

### DIFF
--- a/pkg/account/service.go
+++ b/pkg/account/service.go
@@ -287,7 +287,6 @@ func (a *Service) Reset(data *Account) error {
 // UpsertPrincipalAccess merges principal access to make sure its in sync with expectations
 func (a *Service) UpsertPrincipalAccess(data *Account) error {
 	err := validation.ValidateStruct(data,
-		validation.Field(&data.Status, validation.NotNil, validation.By(isAccountNotLeased)),
 		validation.Field(&data.AdminRoleArn, validation.NotNil),
 		validation.Field(&data.PrincipalRoleArn, validation.NotNil),
 	)


pkg/account/service.go
```diff
@@ -287,7 +287,6 @@ func (a *Service) Reset(data *Account) error {
 // UpsertPrincipalAccess merges principal access to make sure its in sync with expectations
 func (a *Service) UpsertPrincipalAccess(data *Account) error {
 	err := validation.ValidateStruct(data,
-		validation.Field(&data.Status, validation.NotNil, validation.By(isAccountNotLeased)),
 		validation.Field(&data.AdminRoleArn, validation.NotNil),
```
AmudaPalani: Do you still want to validate Status for not nil?

kddejong: I did a little bit of searching I don't see any code in the account manager that uses Status.  So I don't think we need it at all... but double check my work.

AmudaPalani: Since Update principal policy lambda is triggered when new lease is added, do we need to validate Status to ensure it's not nil and it's of Status "Leased"?

kddejong: Its also triggered on account creation.  In that case the account is NotReady.

eschwartz: I'd say this code should only validate for fields that it expects to be in a certain state, in order to do it's job.
So just the role arns, in this case. This change is 👍 to me

kddejong:
Already complete

Issue #137, opened 2019-11-21
Normal users should not be able to create/end leases for other people


joshmarsh:
**Is your feature request related to a problem? Please describe.**
While authenticating with temporary sts credentials mapped to a cognito user called `testuser`, I observed the following behavior.

<pre>
~ dce leases create --budget-amount 100.0 --budget-currency USD --email jane.doe@optum.com <b>--principle-id jdoe99</b>
Lease created: {"accountId":"XXX","budgetAmount":100,"budgetCurrency":"USD","budgetNotificationEmails":["jane.doe@optum.com"],"createdOn":1574347343,"expiresOn":1574952143,"id":"d7586b82-2b57-4ba0-9469-ba59d865e823","lastModifiedOn":1574347343,"leaseStatus":"Active","leaseStatusModifiedOn":1574347343,"leaseStatusReason":"Active","principalId":"<b>jdoe99</b>"}
~ dce leases end --account-id XXX <b>--principle-id jdoe99</b>
Lease ended
</pre>

**Describe the solution you'd like**
Respond with 403 for any leases requests involving a principaID that is not your own.

PR #276, opened 2020-02-07
WIP: Lease auth refactor

eschwartz:
## Proposed changes

- Refactor the /leases/{id}/auth controller to use new patterns for DB access and controller setup.
- Add a new `POST /leases/auth` to login to your active lease (based on your Cognito identity)
- Add functional test coverage for lease auth endpoints (had none, previously)

The new endpoint was trivial to add given the way the new code was structured, so it seemed an easy win to sneak in here.

I don't have any functional test coverage against the Cognito logic. I know @joshmarsh  is working on some functional test support for cognito integration, so I wanted to wait for that to get merged in before implementing the same pattern here.

-----

#### Manual testing

I deployed this PR to a sandbox environment, setup admin and user Cognito Users, and attempted to use the dce CLI to create and login to leases

**As a User by Lease ID**

Can only login to my own lease
```
~/dev/dce/modules$ dce leases list
[
	{
		"accountId": "<redacted>",
		"budgetAmount": 100,
		"budgetCurrency": "USD",
		"budgetNotificationEmails": [
			"a@b.com"
		],
		"createdOn": 1581610356,
		"expiresOn": 1582215156,
		"id": "392e14ab-44fe-4bc8-a813-14c84db58227",
		"lastModifiedOn": 1581610356,
		"leaseStatus": "Active",
		"leaseStatusModifiedOn": 1581610356,
		"leaseStatusReason": "Active",
		"principalId": "testuser"
	},
	{
		"accountId": "<redacted>",
		"budgetAmount": 100,
		"budgetCurrency": "USD",
		"budgetNotificationEmails": [
			"a@b.com"
		],
		"createdOn": 1581610793,
		"expiresOn": 1582215592,
		"id": "08182025-2dc6-4998-af82-8dba2784c463",
		"lastModifiedOn": 1581610793,
		"leaseStatus": "Active",
		"leaseStatusModifiedOn": 1581610793,
		"leaseStatusReason": "Active",
		"principalId": "testadmin"
	}
]

~/dev/dce/modules$ dce leases login -c 392e14ab-44fe-4bc8-a813-14c84db58227
<redacted>

~/dev/dce/modules$ dce leases login -c 08182025-2dc6-4998-af82-8dba2784c463
err:  [POST /leases/{id}/auth][401] postLeasesIdAuthUnauthorized
```

**As an admin user by Lease ID**

Can login to my own lease, and another user's lease

```
~/dev/dce/modules$ dce leases list
[
	{
		"accountId": "<redacted>",
[...]

PR #284, opened 2020-02-17
Feature/user details middleware

joshmarsh:
## Proposed changes

Adds authorization to the leases endpoint.

See https://github.com/Optum/dce/issues/137

## Types of changes

## Checklist


joshmarsh:
There's some overlap between this PR and https://github.com/Optum/dce/pull/276. Let's wait until https://github.com/Optum/dce/pull/276 is merged before merging this one.

cmd/lambda/leases/delete.go
```diff
@@ -14,17 +15,28 @@ import (
...
+	if err != nil {
+		api.WriteAPIErrorResponse(w, err)
+		return
+	}
 
-	lease, err := Services.LeaseService().Delete(leaseID)
+	//If user is not an admin, they can't delete leases for other users
+	if user.Role != api.AdminGroupName && *_lease.PrincipalID != user.Username {
+		log.Printf("User [%s] with role: [%s] attempted to delete a lease for: [%s], but was not authorized", user.Username, user.Role, *_lease.PrincipalID)
+		response.WriteUnauthorizedError(w)
```
kddejong: Should be.
```
err = errors.NewUnathorizedError(...)
api.WriteAPIErrorResponse(w, err)
```

cmd/lambda/leases/main.go
```diff
@@ -73,6 +76,23 @@ type messageBody struct {
...
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		reqCtx, err := muxLambda.GetAPIGatewayContext(r)
+		if err != nil {
+			log.Printf("Failed to parse context object from request: %s", err)
+			api.WriteAPIErrorResponse(w,
+				errors.NewInternalServer("Internal server error", err),
+			)
+			return
+		}
+		user = Services.UserDetailer().GetUser(&reqCtx)
```
kddejong: This is the best way to do this?  Seems weird to store something outside of the middleware?  Is this function that should be in the API?

pkg/api/response/error.go
```diff
@@ -206,6 +206,16 @@ func WriteBadRequestError(w http.ResponseWriter, message string) {
 	)
 }
 
+// WriteUnauthorizedError - Writes the unauthorized error.
+func WriteUnauthorizedError(w http.ResponseWriter) {
```
kddejong: Move this to the errors pkg please

cmd/lambda/leases/main.go
```diff
@@ -64,6 +66,7 @@ var (
 	baseRequest              url.URL
 	//cognitoUserPoolId        string
 	//cognitoAdminName         string
+	user *api.User
```
kddejong: Can we put the validate functionality on the struct itself?  Then we can just call a function on that struct?

cmd/lambda/leases/main.go
joshmarsh: Yeah, it's kind of weird. I was trying to avoid using a context based on some discussions we'd had. But I think this is a good use case for it.

PR #289, opened 2020-02-18
Fix Account status transition

eschwartz:
## Proposed changes

We had an error in our system, that when a lease was deactivated, the associated account would not transition from `Leased` --> `Inactive`. This turned out to be a bug in the `AccountService.Update()` method, that it would not actually persist updates to the DB.

Functional tests now poll the `GET /accounts` API to verify status changes. As a bonus, this allows us to entirely remove direct `dbSvc` calls from our functional tests. Plus, we now have E2E coverage for the account reset process (minus aws-nuke itself, which is disabled in our test environment)

In order to get these tests working E2E through the reset process, I had to modify how the reset codebuild runs, when `var.reset_nuke_toggle = false`. Previously, this setting would run aws-nuke the dry run mode. I changed it skip calling aws-nuke at all. The issue was that failures in the aws-nuke dry run would make the entire reset process fail (for example, when we try to nuke the master account, as we do in our functional tests), and result in accounts stuck in `status=NotReady`. Besides getting our tests passing, I believe this is a better experience for end users who don't want to run aws-nuke at all. And it just makes sense that if `reset_nuke_toggle = false`, then reset won't run.

Also:
- Fix error in pipelines, when recreating S3 tfstate buckets with the same name as had previously been used. (use the lease ID instead, which is predictable, but unique per lease)
- Do not reinstall gosec every time you run `make lint` (was making linting painful to run locally)

## Types of changes

## Checklist

### DIFF
--- a/cmd/lambda/accounts/list.go
+++ b/cmd/lambda/accounts/list.go
@@ -19,7 +19,7 @@ func GetAccounts(w http.ResponseWriter, r *http.Request) {
 	query := &account.Account{}
 	err := decoder.Decode(query, r.URL.Query())
 	if err != nil {
-		response.WriteRequestValidationError(w, fmt.Sprintf("Error parsing query params"))
+		response.WriteRequestValidationError(w, fmt.Sprintf("Error parsing query params: %s", err))
 		return
 	}
 
--- a/cmd/lambda/accounts/update_test.go
+++ b/cmd/lambda/accounts/update_test.go
@@ -46,7 +46,7 @@ func TestUpdateAccountByID(t *testing.T) {
 		{
 			name:      "success",
 			accountID: "123456789012",
-			reqBody:   fmt.Sprintf("{\"metadata\": {\"key\": \"value\"}}"),
+			reqBody:   "{\"metadata\": {\"key\": \"value\"}}",
 			reqAccount: &account.Account{
 				Metadata: map[string]interface{}{
 					"key": "value",
@@ -72,7 +72,7 @@ func TestUpdateAccountByID(t *testing.T) {
 		{
 			name:       "failure validation",
 			accountID:  "123456789012",
-			reqBody:    fmt.Sprintf("{\"accountStatus\": \"NotReady\""),
+			reqBody:    "{\"accountStatus\": \"NotReady\"",
 			reqAccount: &account.Account{},
 			expResp: response{
 				StatusCode: 400,
@@ -84,7 +84,7 @@ func TestUpdateAccountByID(t *testing.T) {
 		{
 			name:      "failure db",
 			accountID: "123456789012",
-			reqBody:   fmt.Sprintf("{\"metadata\": {\"key\": \"value\"}}"),
+			reqBody:   "{\"metadata\": {\"key\": \"value\"}}",
 			reqAccount: &account.Account{
 				Metadata: map[string]interface{}{
 					"key": "value",
@@ -100,7 +100,7 @@ func TestUpdateAccountByID(t *testing.T) {
 		{
 			name:      "failure decode",
 			accountID: "123456789012",
-			reqBody:   fmt.Sprintf("{\"metadata\": \"key\": \"value\"}}"),
+			reqBody:   "{\"metadata\": \"key\": \"value\"}}",
 			reqAccount: &account.Account{
 				Metadata: map[string]interface{}{
 					"key": "value",
--- a/cmd/lambda/publish_lease_events/main.go
+++ b/cmd/lambda/publish_lease_events/main.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"encoding/json"
 	"errors"
-	"fmt"
 	"log"
 
 	"github.com/Optum/dce/pkg/common"
@@ -12,7 +11,6 @@ import (
 	errors2 "github.com/Optum/dce/pkg/errors"
 	"github.com/aws/aws-lambda-go/events"
 	"github.com/aws/aws-lambda-go/lambda"
-	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/session"
 	"github.com/aws/aws-sdk-go/service/dynamodb"
 	"github.com/aws/aws-sdk-go/service/dynamodb/dynamodbattribute"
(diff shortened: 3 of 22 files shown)


tests/acceptance/api_test.go
```diff
@@ -2173,12 +2224,21 @@ func apiRequest(t *testing.T, input *apiRequestInput) *apiResponse {
 	return apiResp
 }
 
-func parseResponseJSON(t *testing.T, resp *apiResponse) map[string]interface{} {
+func parseResponseJSON(t require.TestingT, resp *apiResponse) map[string]interface{} {
```
eschwartz: changed the interface here, so we can use these methods in the context of  `assert`, `require`, and `retry` calls

tests/acceptance/testutil/retry.go
```diff
@@ -62,6 +62,10 @@ func (r *R) Fail() {
 	r.failed = true
 }
 
+func (r *R) FailNow() {
```
eschwartz: to satisfy `require.TestingT`, giving us more flexibility in reusing testing methods for that interface

tests/acceptance/api_test.go
```diff
@@ -919,12 +936,26 @@ func TestApi(t *testing.T) {
 				require.Nil(t, err)
 				require.NotNil(t, resJSON["leaseStatusModifiedOn"])
 
-				// Check the lease is in the DB
-				// (since we dont' yet have a GET /leases endpoint
```
eschwartz: > since we dont' yet have a GET /leases endpoint

😮

cmd/lambda/publish_lease_events/main.go
```diff
@@ -94,38 +94,32 @@ func handleRecord(input *handleRecordInput) error {
 		prevLeaseStatus := prevLeaseStatusAttr.String()
 
 		// Grab the new lease status
-		nextLeaseStatusAttr, ok := record.Change.NewImage["LeaseStatus"]
+		newLeaseStatusAttr, ok := record.Change.NewImage["LeaseStatus"]
```
eschwartz: just a var name change -- I got confused, thinking that the `nextLeaseStatus` had not yet been applied.

cmd/lambda/publish_lease_events/main.go
```diff
@@ -94,38 +94,32 @@ func handleRecord(input *handleRecordInput) error {
...
-			// Before adding the account to any queues, make sure the account is
-			// updated to NotReady state.
-			var acct *db.Account
-			acct, err := input.dbSvc.TransitionAccountStatus(lease.AccountID, db.Leased, db.NotReady)
+			// Set account status to NotReady
+			acct, err := input.dbSvc.GetAccount(lease.AccountID)
 			if err != nil {
-				log.Printf("ERROR: Failed to mark AccountStatus=NotReady for %s, after lease for %s became inactive",
-					lease.AccountID, lease.PrincipalID)
+				return fmt.Errorf("ERROR: Failed to retrieve account %s: %s", lease.AccountID, err)
```
kddejong: we took this out and back in.  The concern here is how dynamodb streams and Lambdas work.  Anything that can cause the Lambda to error will result in re-attempted processing of that message until its clear.  These default values are large.

Also can you change this to errors.InternalServerError.  Those errors don't just have to be used for http status errors that's just a side benefit.

eschwartz: Yep, good point. Let's talk through the situation here in more detail. I think there's a better solution to be had.

cmd/lambda/accounts/list.go
```diff
@@ -19,7 +19,7 @@ func GetAccounts(w http.ResponseWriter, r *http.Request) {
 	query := &account.Account{}
 	err := decoder.Decode(query, r.URL.Query())
 	if err != nil {
-		response.WriteRequestValidationError(w, fmt.Sprintf("Error parsing query params"))
+		response.WriteRequestValidationError(w, fmt.Sprintf("Error parsing query params: %s", err))
```
eschwartz: These calls to `fmt.Sprintf` without formatting params were causing linting errors.

cmd/lambda/accounts/update_test.go
```diff
@@ -46,7 +46,7 @@ func TestUpdateAccountByID(t *testing.T) {
 		{
 			name:      "success",
 			accountID: "123456789012",
-			reqBody:   fmt.Sprintf("{\"metadata\": {\"key\": \"value\"}}"),
```
eschwartz: linting errors

pkg/account/service.go
```diff
@@ -124,7 +124,7 @@ func (a *Service) Update(ID string, data *Account) (*Account, error) {
 		return nil, err
 	}
 
-	err = mergo.Merge(account, *data)
+	err = mergo.Merge(account, *data, mergo.WithOverride)
```
eschwartz: This was the whole bug right here. `mergo.Merge()` does not override values in the src struct by default.

pkg/account/service_test.go
```diff
@@ -197,6 +197,30 @@ func TestUpdate(t *testing.T) {
 			},
 			returnErr: nil,
 		},
+		{
```
eschwartz: unit test coverage for `mergo.Merge` changes (verified this test fails, without those merge changes)

scripts/install_ci.sh
```diff
@@ -15,3 +15,5 @@ go get github.com/axw/gocov/gocov
 go get github.com/AlekSi/gocov-xml
 go get github.com/matm/gocov-html
 go get -u github.com/golangci/golangci-lint/cmd/golangci-lint
+
+go get github.com/securego/gosec/cmd/gosec
```
eschwartz: install gosec from Azure pipeline, instead of from lint.sh, to be more friendly to running `make lint` locally

tests/acceptance/api_test.go
```diff
@@ -397,17 +398,17 @@ func TestApi(t *testing.T) {
 			require.NotNil(t, data["lastModifiedOn"])
 			require.NotNil(t, data["leaseStatusModifiedOn"])
 
-			// Create the Decommission Request Body
-			body = leaseRequest{
-				PrincipalID: principalID,
-				AccountID:   acctID,
-			}
+			// Account should be marked as status=Leased
+			waitForAccountStatus(t, apiURL, acctID, "Leased")
```
eschwartz: new test coverage, to verify account status update

Added a `waitForAccountStatus` helper util, since this is a common assertion in our tests

tests/acceptance/api_test.go
```diff
@@ -1036,10 +1039,8 @@ func TestApi(t *testing.T) {
 			})
 
 			t.Run("STEP: Create Lease", func(t *testing.T) {
-				// Account is being reset, so it's not marked as "Ready".
-				// Update the DB to be ready, so we can create a lease
-				_, err := dbSvc.TransitionAccountStatus(accountID, db.NotReady, db.Ready)
```
eschwartz: 👋 `dbSvc` 👋

tests/acceptance/api_test.go
```diff
@@ -1306,6 +1336,77 @@ func TestApi(t *testing.T) {
 		}, getResJSON["metadata"])
 	})
 
+	t.Run("Create multiple leases on one account", func(t *testing.T) {
```
eschwartz: We specifically saw this bug when recreating a lease against the same account, multiple times. 

I did try recreating a lease 20 times while testing this out locally. It passed, but took some 15m to run. Generally I saw an error by attempt no 3. Still takes a couple minutes to run, but I think it's a worthwhile test to have in the arsenal.

tests/acceptance/api_test.go
```diff
@@ -2415,8 +2525,14 @@ func createAdminRole(t *testing.T, awsSession client.ConfigProvider, adminRoleNa
 		})
 		require.Nil(t, err)
 	}
-	// IAM Role takes a while to propagate....
-	//time.Sleep(10 * time.Second)
+
+	// Wait for the role to be assumable
+	log.Println("Created admin test role. Waiting for role to be assumeable")
+	testutil.Retry(t, 30, time.Millisecond*500, func(r *testutil.R) {
```
eschwartz: Turns out a lot of our `apiRequest` retries were only necessary because our IAM role wasn't yet ready to use.
Waiting for this to be ready upfront reduces some noise in tests.

tests/acceptance/api_test.go
```diff
@@ -355,27 +356,34 @@ func TestApi(t *testing.T) {
...
 
-			// Create an Account Entry
-			acctID := "123"
-			principalID := "user"
-			timeNow := time.Now().Unix()
-			err := dbSvc.PutAccount(db.Account{
-				ID:             acctID,
-				AccountStatus:  db.Ready,
-				LastModifiedOn: timeNow,
+			// Add the current account to the account pool
```
eschwartz: Trying to avoid direct DB calls whereever possible.

tests/acceptance/api_test.go
```diff
@@ -708,61 +722,6 @@ func TestApi(t *testing.T) {
 
 		})
 
-		t.Run("Should not be able to create lease if there's already an existing lease for the principal", func(t *testing.T) {
```
eschwartz: I moved this test to an existing test flow, so we don't need to recreate the account and all that.

tests/acceptance/api_test.go
```diff
@@ -1136,72 +1141,8 @@ func TestApi(t *testing.T) {
 					assert.Equal(t, 1, len(results), "one lease should be returned")
 					assert.Equal(t, "Inactive", results[0]["leaseStatus"])
 
-					t.Run("STEP: Recreate lease against same account", func(t *testing.T) {
```
eschwartz: This is now covered in a dedicated test, that attempts to create/destroy/re-create leases several times.

PR #311, opened 2020-03-06
Validate that if an account doesn't exist we skip

kddejong:
## Proposed changes

- Some cleanup around the reset queue process

## Types of changes

## Checklist


cmd/lambda/publish_lease_events/main.go
```diff
@@ -126,6 +126,10 @@ func handleRecord(input *handleRecordInput) error {
 					log.Printf("ERROR: Failed to mark get account %q", lease.AccountID)
 					return err
 				}
+				if acct == nil {
```
kddejong: with functional testing we are occasionally throwing in `nil` values to the queue.  This is because we are truncating the table and not able to query the account.

tests/acceptance/api_test.go
```diff
@@ -357,6 +358,8 @@ func TestApi(t *testing.T) {
 	t.Run("Lease Creation and Deletion", func(t *testing.T) {
 
 		t.Run("Should be able to create and destroy a lease", func(t *testing.T) {
+			truncateAccountTable(t, dbSvc)
```
kddejong: Cleanup the tables before we start lease work

modules/reset.tf
```diff
@@ -2,7 +2,18 @@
 resource "aws_sqs_queue" "account_reset" {
   name                       = "account-reset-${var.namespace}"
   tags                       = var.global_tags
-  visibility_timeout_seconds = 30
+  visibility_timeout_seconds = 60
+  redrive_policy = jsonencode({
```
kddejong: Create a queue redrive policy.  This will move messages that are causing issues into a dead letter queue if they are failing.

pkg/common/storager.go
```diff
@@ -108,7 +109,12 @@ func (stor S3) Download(bucket string, key string, filepath string) error {
 	if err != nil {
 		return err
 	}
-	defer file.Close()
+	defer func() {
```
kddejong: go sec issue

tests/acceptance/api_test.go
```diff
@@ -448,12 +451,15 @@ func TestApi(t *testing.T) {
...
 			timeNow := time.Now().Unix()
 			err := dbSvc.PutAccount(db.Account{
-				ID:             leasedAccountID,
-				AccountStatus:  db.Ready,
-				LastModifiedOn: timeNow,
+				ID:                 leasedAccountID,
+				AccountStatus:      db.Ready,
+				AdminRoleArn:       fmt.Sprintf("arn:aws:iam::%s:role/adminRole", leasedAccountID),
+				PrincipalRoleArn:   fmt.Sprintf("arn:aws:iam::%s:role/principalRole", leasedAccountID),
+				PrincipalPolicyArn: fmt.Sprintf("arn:aws:iam::%s:policy/principalPolicy", leasedAccountID),
```
kddejong: Because we are loading records directly into the system and not using the APIs I did a little more work to clean up the records to match expectations.  This is a short term fix to a larger problem.

tests/acceptance/db_test.go
```diff
@@ -1,793 +1,15 @@
 package tests
```
kddejong: A lot of this code is no longer used.  It exists only because it is used in the test scripts.  Removing the tests as it is deprecated and adds to the issues to the testing system.

tests/acceptance/api_test.go
```diff
@@ -448,12 +451,14 @@ func TestApi(t *testing.T) {
...
 			timeNow := time.Now().Unix()
 			err := dbSvc.PutAccount(db.Account{
-				ID:             leasedAccountID,
-				AccountStatus:  db.Ready,
-				LastModifiedOn: timeNow,
+				ID:               leasedAccountID,
+				AccountStatus:    db.Ready,
+				AdminRoleArn:     fmt.Sprintf("arn:aws:iam::%s:role/adminRole", leasedAccountID),
+				PrincipalRoleArn: fmt.Sprintf("arn:aws:iam::%s:role/principalRole", leasedAccountID),
+				LastModifiedOn:   timeNow,
```
kddejong: Because we are loading records directly into the system and not using the APIs I did a little more work to clean up the records to match expectations. This is a short term fix to a larger problem.

cmd/lambda/publish_lease_events/main.go
kddejong: So if we still can't get the account at this point lets assume its gone and we can skip the rest of the work here.

PR #318, opened 2020-03-18
Fix to allow budget notification email templates > 4Kb

eschwartz:
## Proposed changes

When configuring email templates > 4kb in tfvars,
terraform apply would fail with:

> Error: Error creating Lambda function: InvalidParameterValueException: Lambda was unable to configure your environment variables because the environment variables you have provided exceeded the 4KB limit.

This fix moves the templates to S3 object (instead of env vars) so we are not bound by
the 4kb env var limitation

## Types of changes

## Checklist

### DIFF
--- a/cmd/lambda/update_lease_status/main.go
+++ b/cmd/lambda/update_lease_status/main.go
@@ -3,6 +3,8 @@ package main
 import (
 	"encoding/json"
 	"fmt"
+	"github.com/aws/aws-sdk-go/service/s3"
+	"github.com/aws/aws-sdk-go/service/s3/s3manager"
 	"log"
 	"time"
 
@@ -54,6 +56,12 @@ func main() {
 			log.Fatalf("Failed to configure Usage service %s", err)
 		}
 
+		// Configure the S3 service
+		s3Svc := &common.S3{
+			Client:  s3.New(awsSession),
+			Manager: s3manager.NewDownloader(awsSession),
+		}
+
 		err = lambdaHandler(&lambdaHandlerInput{
 			dbSvc:                                  dbSvc,
 			lease:                                  lease,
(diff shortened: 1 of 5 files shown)


cmd/lambda/update_lease_status/main.go
```diff
@@ -54,6 +56,12 @@ func main() {
 			log.Fatalf("Failed to configure Usage service %s", err)
 		}
 
+		// Configure the S3 service
```
eschwartz: Of course, it would be nice to use the new configuration service and all that,
but I I didn't want to rabbit-hole on it.

cmd/lambda/update_lease_status/main.go
```diff
@@ -65,10 +73,12 @@ func main() {
 			snsSvc:                                 &common.SNS{Client: sns.New(awsSession)},
 			leaseLockedTopicArn:                    common.RequireEnv("LEASE_LOCKED_TOPIC_ARN"),
 			emailSvc:                               &email.SESEmailService{SES: ses.New(awsSession)},
+			s3Svc:                                  s3Svc,
 			budgetNotificationFromEmail:            common.RequireEnv("BUDGET_NOTIFICATION_FROM_EMAIL"),
 			budgetNotificationBCCEmails:            common.RequireEnvStringSlice("BUDGET_NOTIFICATION_BCC_EMAILS", ","),
-			budgetNotificationTemplateHTML:         common.RequireEnv("BUDGET_NOTIFICATION_TEMPLATE_HTML"),
-			budgetNotificationTemplateText:         common.RequireEnv("BUDGET_NOTIFICATION_TEMPLATE_TEXT"),
+			budgetNotificationTemplatesBucket:      common.RequireEnv("BUDGET_NOTIFICATION_TEMPLATES_BUCKET"),
```
eschwartz: Main change is the env var configuration here -- we're pointing at S3 objects, instead of embedding the templates in-full as env vars. We then just need to load the templates from S3 before sending emails.

modules/update_lease_status.tf
```diff
@@ -49,6 +54,19 @@ module "update_lease_status_lambda" {
   }
 }
 
+// Upload budget notification email templates to S3
```
eschwartz: We do have some tfvars where we explicitly ask the user to provide s3 buckets and keys, and expect the user to upload the file themselves (eg. `var.reset_nuke_template_bucket` / `var.reset_nuke_template_key`). 

But changing the tfvars to accept s3 buckets/keys would be a breaking change, so I'd prefer to avoid it.

PR #319, opened 2020-03-18
Feature/add supported services

nathanagood:
## Proposed changes

Ran the tool for generating sample policy and markdown documentation per [the README](tools/awsnukedocgen/README.md) to generate the new policy and updated the one in place for the template.

## Types of changes

## Checklist


modules/fixtures/policies/principal_policy.tmpl
```diff
@@ -91,7 +91,6 @@
         "codestar:*",
         "cognito-identity:*",
         "cognito-idp:*",
-        "cognito-sync:*",
```
eschwartz: @nathanagood what's the story behind removing some of the services in this policy?

nathanagood: Cognito Sync doesn't appear anywhere in the AWS Nuke resources code. SimpleDB does, but it's an old, deprecated service and use of it should be avoided compared to other services (DDB).

eschwartz:
I'm all good with this, whenever we can get the build passing. thanks @nathanagood !

PR #331, opened 2020-04-06
Update lease should update LeaseStatusModifiedOn

AmudaPalani:
## Proposed changes

Update lease should update LeaseStatusModifiedOn

## Types of changes

## Checklist


tests/acceptance/api_test.go
```diff
@@ -1330,6 +1330,19 @@ func TestApi(t *testing.T) {
...
+				method: "GET",
+				url:    apiURL + fmt.Sprintf("/leases?principalId=test-user&accountId=%s", accountID),
+				json:   nil,
+			})
+
+			results := parseResponseArrayJSON(t, resp)
+			assert.Equal(t, 200, resp.StatusCode)
+			assert.Equal(t, 1, len(results), "one lease should be returned")
+			assert.Equal(t, "Active", results[0]["leaseStatus"])
+			assert.NotNil(t, results[0]["leaseStatusModifiedOn"])
```
nathanagood, line 1344: We're probably not doing this anywhere else, but do we want to add any different checks other than NotNil? Should it be formatted a certain way? Greater than today, etc.?

AmudaPalani: @nathanagood I will add a check to verify leaseStatusModifiedOn date is equal to CreatedOn date.

pkg/lease/service.go
```diff
@@ -91,6 +91,7 @@ func (a *Service) Save(data *Lease) error {
 	} else {
 		lastModifiedOn = data.LastModifiedOn
 		data.LastModifiedOn = &now
+		data.StatusModifiedOn = &now
```
eschwartz, line 94: @AmudaPalani I don't think this is quite what we want.

The `LeaseStatusModifiedOn` field should only be updated if the `LeaseStatus` field changes.
Unfortunately, we have no way within the `Save()` method to know if the `LeaseStatus` field has been updated (we only have the new lease object, not the old one).

I'm not sure how best to approach this --we may need to set this timestamp  in some of the higher-level service methods (eg. `Create()` / `Delete()`). What do you think?

AmudaPalani: This line 92-94 will be executed in two scenarios only:
a) Update a lease status from active to inactive (Decommission).
b) Update a lease status from inactive to active (Same account getting leased).
In both these scenarios, we want to update LeaseStatusModifiedOn.

PR #467, opened 2023-11-14
Switch from deny and allow config, to only deny and allow everything …

bjfish25:
## Proposed changes

Previously we have been micro-managing the allow list for DCE. Given there is better solutions for this in-house, and to make future use-cases easier on new AWS services, it would be easier to just have a blanket allowed, only limited by some basic Deny's already present

This pr
- removes specific allow permissions
- add's blanket default allow
- upgrades terraform to 0.13.7

## Types of changes

## Checklist


dilip0515 (approved):
lgtm

PR #479, opened 2024-02-23
Update db.go

mxu88:
## Proposed changes

- Get rid of default limit of 25 when querying for leases

## Types of changes

## Checklist

### DIFF
--- a/pkg/db/db.go
+++ b/pkg/db/db.go
@@ -625,18 +625,16 @@ type GetLeasesOutput struct {
 
 // GetLeases takes a set of filtering criteria and scans the Leases table for the matching records.
 func (db *DB) GetLeases(input GetLeasesInput) (GetLeasesOutput, error) {
-	var limit int64
 	filters := make([]string, 0)
 	filterValues := make(map[string]*dynamodb.AttributeValue)
 
-	if input.Limit > 0 {
-		limit = input.Limit
-	}
-
 	scanInput := &dynamodb.ScanInput{
 		TableName:      aws.String(db.LeaseTableName),
-		Limit:          &limit,
 		ConsistentRead: aws.Bool(db.ConsistentRead),
+	}
+
+	if input.Limit > 0 {
+		scanInput.Limit = &input.limit
 	}
 
 	// Build the filter clauses.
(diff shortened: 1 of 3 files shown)


dilip0515 (approved):

pkg/db/db.go
```diff
@@ -625,7 +625,7 @@ type GetLeasesOutput struct {
 
 // GetLeases takes a set of filtering criteria and scans the Leases table for the matching records.
 func (db *DB) GetLeases(input GetLeasesInput) (GetLeasesOutput, error) {
-	limit := int64(25)
+	var limit int64
```
bjfish25, line 628: Isn't this setting the default limit to 0? 

According to [docs](https://docs.aws.amazon.com/sdk-for-go/api/service/dynamodb/#ScanInput) without it will stop at 1 megabyte regardless

mxu88: Are you saying that setting the limit to 0 will cause an error, or that we may need to consider adding code in the case that we hit the 1 megabyte limit? I'm also seeing that there is `min:1` do you know if that will cause any issue if we're setting limit to 0?
Screenshot 2024-02-26 at 9 37 29 AM

bjfish25: I don't know exactly how it is handled, a set 0 could be such that it is considered unset, but I would probably expect it to error with an invalid input.

I think what you could do here is ensure Limit is unset in the ScanInput in the case that `if input.Limit > 0 {`, and it will always try to return the max 1 mb of data

mxu88: @bjfish25 I made a change for the limit. Please check again

bjfish25 (approved):

pkg/db/db.go
```diff
@@ -625,20 +625,18 @@ type GetLeasesOutput struct {
...
-	}
-
 	scanInput := &dynamodb.ScanInput{
 		TableName:      aws.String(db.LeaseTableName),
-		Limit:          &limit,
 		ConsistentRead: aws.Bool(db.ConsistentRead),
 	}
 
+	if input.Limit > 0 {
+		scanInput.Limit = &input.limit
```
bjfish25, line 637: Actually this doesn't look like this works with the reference

bjfish25 (changes requested):
@mxu88 can you view the builds in Azure pipelines? https://dev.azure.com/OptumDCE/DCE/_build/results?buildId=2982&view=logs&j=a66b8d62-1279-5d41-9337-b72a91aa4761&t=87d4aa90-47d8-5c5e-276c-0d2858f5cd25

pkg/db/db.go
mxu88: Oops had a letter casing error haha... The build is successful now. @dilip0515 is also getting me access to the devops

dilip0515 (approved):
lgtm

PR #506, opened 2024-11-18
Update  use Query instead of Scan

dilip0515:
## Proposed changes

- [x] Update  use Query instead of Scan for Get Leases from Lease table

## Types of changes

## Checklist


pkg/db/db.go
```diff
@@ -655,18 +655,18 @@ func (db *DB) GetLeases(input GetLeasesInput) (GetLeasesOutput, error) {
...
-		scanInput.ExclusiveStartKey = make(map[string]*dynamodb.AttributeValue)
+		queryInput.ExclusiveStartKey = make(map[string]*dynamodb.AttributeValue)
 		for k, v := range input.StartKeys {
-			scanInput.ExclusiveStartKey[k] = &dynamodb.AttributeValue{S: aws.String(v)}
+			queryInput.ExclusiveStartKey[k] = &dynamodb.AttributeValue{S: aws.String(v)}
 		}
 	}
 
-	output, err := db.Client.Scan(scanInput)
+	output, err := db.Client.Query(queryInput)
```
bjfish25, line 669: Can we only do the query if an index field is present, and a scan otherwise?

pkg/db/db.go
```diff
@@ -628,13 +628,16 @@ func (db *DB) GetLeases(input GetLeasesInput) (GetLeasesOutput, error) {
 	filters := make([]string, 0)
 	filterValues := make(map[string]*dynamodb.AttributeValue)
 
+	queryInput := &dynamodb.QueryInput{
+		TableName:      aws.String(db.LeaseTableName),
+		ConsistentRead: aws.Bool(db.ConsistentRead),
```
bjfish25, line 633: Would this work given that no index is given?

I think the easist way to accomplish this is
1. Go back to largely how it was
2. Make a variable called, queryable, and queryIndex (or something like that) and mark it false to start
3. If one of the index filters are set, set the field (make sure not to overwrite, we probably want to prioritize principal_id)
4. Make the query with filters if queryable, make the scan with filters otherwise
5. Output the results to the same spot

dilip0515:
@bjfish25  - can you review this PR? updated per the review comments

pkg/db/db.go
```diff
@@ -625,74 +625,124 @@ type GetLeasesOutput struct {
...
 	}
 
-	// Build the filter clauses.
-	if input.Status != "" {
-		filters = append(filters, "LeaseStatus = :status")
-		filterValues[":status"] = &dynamodb.AttributeValue{S: aws.String(string(input.Status))}
+	// Check if the PrincipalId index exists
+	var queryIndex bool
+	for _, index := range tableDesc.Table.GlobalSecondaryIndexes {
+		if *index.IndexName == "PrincipalId" {
```
bjfish25, line 641:
```suggestion
		if *index.IndexName == "PrincipalId" && input.PrincipalID != "" {
```

bjfish25 (approved):
LGTM

dilip0515:
having issues to merge opened a new PR 508

